PR #8, opened 2017-07-09
removed starting of service for fedora and centos

andrewhsu:
Running `install.sh` inside docker container, receiving error message:
```
Failed to get D-Bus connection: Operation not permitted
```
Even though the rpm is successfully installed.

In rpm world, should not start systemd service after install of rpm
package if the service is not already started before install, which is
the case for first install of the package. From fedora [docs](https://fedoraproject.org/wiki/Packaging:Systemd#Why_don.27t_we....).

Signed-off-by: Andrew Hsu <andrewhsu@docker.com>


andrewhsu:
Also added another commit to check return value of `install.sh` in the `verify-docker-install` script.

PR #13, opened 2017-07-18
Make yum repo point to correct lsb_dist

seemethere:
`yum_repo` mistakenly was hardcoded to `centos` now it should be based
off of `$lsb_dist`


andrewhsu:
LGTM

PR #17, opened 2017-07-26
WIP: Update to include support map

seemethere:
This makes it more explicit which distros, architectures, and distro
versions we support.

TODO: Add support for detecting potential EE support and redirect to the
EE page


seemethere:
Closing, will be submitting a larger PR later.

PR #23, opened 2017-08-11
Use mirror provided by AzureChinaCloud

zhang-hua:
In order to install docker engine on Ubuntu from China mirror provided by AzureChinaCloud  if mirror option is present.

usage:
sh install --mirror AzureChinaCloud

Without this change, the docker install will always fail on downloading from global site due to network issue.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -75,6 +75,9 @@ case "$mirror" in
 	Aliyun)
 		DOWNLOAD_URL="https://mirrors.aliyun.com/docker-ce"
 		;;
+	AzureChinaCloud)
+		DOWNLOAD_URL="https://mirror.azure.cn/docker-ce"
+		;;
 esac
 
 command_exists() {


seemethere:
So as far as this script goes, this should only support `docker-ce` installation from this point forward and not `docker-engine`.

If this PR were to get in you'd expect vastly different behavior depending on which mirror you are using.

Is it possible to have the `AzureChinaCloud` mirror, mirror the current repo @ download.docker.com?

seemethere:
The ideal change would be to just add another option in the mirror section like: 
https://github.com/docker/docker-install/commit/cfa611a019a58c09a1528be239c273225aec92af#diff-3fbb47e318cd8802bd325e7da9aaabe8R75

zhang-hua:
will investigate what you have asked and update this PR if possible.

zhang-hua:
We have updated AzureChinaCloud mirror 'https://mirror.azure.cn/docker-ce/' to support 'docker-ce' installation. please review this PR. thanks!

hadifarnoud:
it doesn't actually work. I tried `--mirror Aliyun`

I had to change `DEAFULT_DOWNLOAD_URL` to `https://mirrors.aliyun.com/docker-ce` and even that still not working

PR #27, opened 2017-08-29
Adds support for raspbian stretch

seemethere:
## Overview

Basically uses the jessie repo for raspbian stretch due to raspbian
jessie download links being taken off the official raspbian site.

## NOTE
This is only a temporary fix and should not be considered permanent

ping @StefanScherer @alexellis


StefanScherer:
Excellent! That will make many Raspbian users happy to get started with Docker 17.05 very easy.

PR #47, opened 2017-11-13
Add debian buster to install script

seemethere:
Debian Buster builds for Docker CE are now available on the test channel.

This PR adds:
* Support in the install script
* Support in the Makefile checks
* Support in the `Jenkinsfile`

Relates to: https://github.com/docker/for-linux/issues/102

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,6 +7,7 @@ def verifyTargets = [
   'x86_64-verify-install-debian-wheezy',
   'x86_64-verify-install-debian-jessie',
   'x86_64-verify-install-debian-stretch',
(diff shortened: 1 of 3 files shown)


thaJeztah:
Off-topic: Wondering; should we keep a list of "no longer supported" distro versions, so that we can print an explicit message that no updates will be published, but old versions can still be installed manually?

seemethere:
@thaJeztah That might be good, but I think it'd be better served in the docs.

cc @friism @mistyhacks

mistyhacks:
This kind of thing would be almost impossible to keep up to date in docs...

Issue #43, opened 2017-10-18
Could python be removed to reduce the install footprint?


michaelbarton:
Please excuse my ignorance if this is a stupid issue request. My understanding is that docker is written in go, so I am curious as to why python installed by this script? I'm not using python so having it installed via `python-software-properties` or `software-properties-common` to use docker feels a little superfluous.

I've tried to google search if docker has any python dependencies before opening this issue, but I couldn't see any.

seemethere:
Yup I'm not entirely sure what it's being used for either. Removed it in #51 

Closed with #51

PR #51, opened 2017-12-02
Remove software-properties-common

seemethere:
Doesn't seem to be needed anymore

Relates to #43 


seemethere:
Closes #43

PR #57, opened 2018-01-24
Add Parrot support

PalinuroSec:
even if the latest versions of Parrot OS are based on Debian buster, to consider Parrot as based on Stretch is completely safe in this script, and it works on all the computers we tested it on.


seemethere:
Closing due to inactivity

PR #60, opened 2018-01-24
Add deprecation notices for zesty and f25

seemethere:
Both distributions have reached their EOL and will no longer be supported in Docker CE.

This adds a way to remove them from the script while giving proper warning to affected users.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -105,6 +105,17 @@ is_dry_run() {
 	fi
 }
 
+deprecation_notice() {
+	distro=$1
+	date=$2
+	echo
+	echo "DEPRECATION WARNING:"
+	echo "    The distribution, $distro, will no longer be supported in this script as of $date."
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -374,6 +385,13 @@ do_install() {
 					add_debian_backport_repo "$dist_version"
 				fi
 			fi
+
+			# TODO: March 1, 2018 delete from here,
+			if [ "$lsb_dist" =  "ubuntu" ] && [ "$dist_version" = "zesty" ]; then
+				deprecation_notice "$lsb_dist $dist_version" "March 1, 2018"
+			fi
+			# TODO: March 1, 2018 delete to here,
```
jose-bigio: Are the TODOs addressed? 
Do you need them anymore?

seemethere: The TODO's are for the future so we remember to delete these lines when the time comes

PR #63, opened 2018-02-27
Allow overwriteable download url

seemethere:


seemethere:
@jose-bigio we could do that if this was a `bash` script but unfortunately it's a `sh` script. Wish we could use some `bashism`s

PR #64, opened 2018-03-03
Remove Fedora 25 and Ubuntu Zesty since they are end of life

jhult:


seemethere:
Can you remove the merge commit?

seemethere:
Have no idea why `fedora-25` keeps getting included in the testing cycles but I'll overlook that failure.

LGTM

PR #69, opened 2018-04-04
Add ability to specify custom repository file

seemethere:

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -33,6 +33,11 @@ if [ -z "$DOWNLOAD_URL" ]; then
 	DOWNLOAD_URL=$DEFAULT_DOWNLOAD_URL
 fi
 
+DEFAULT_REPO_FILE="docker-ce.repo"
+if [ -z "$REPO_FILE" ]; then
+	REPO_FILE="$DEFAULT_REPO_FILE"
+fi
+
 SUPPORT_MAP="
 x86_64-centos-7
 x86_64-fedora-26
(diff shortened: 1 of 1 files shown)


jose-bigio:
Hey let me know when it's ready for review, I see you are still adding changes.

seemethere:
Should work now:

```
~/work/docker-install custom_repo ubuntu@davenport
❯ docker-here fedora:27
[root@f81c2a005ccb v]# REPO_FILE=bleh-bleh.repo sh install.sh
# Executing docker install script, commit: UNKNOWN
Error: Unable to curl repository file https://download.docker.com/linux/fedora/bleh-bleh.repo, is it valid?
[root@f81c2a005ccb v]# exit
```

install.sh
```diff
@@ -33,6 +33,11 @@ if [ -z "$DOWNLOAD_URL" ]; then
 	DOWNLOAD_URL=$DEFAULT_DOWNLOAD_URL
 fi
 
+DEFAULT_REPO_FILE="docker-ce.repo"
+if [ -z "$REPO_FILE" ]; then
+	REPO_FILE="$DEFAULT_REPO_FILE"
```
jose-bigio: Personally I would set REPO_FILE="docker-ce.repo" in the if statement, and not create an extra variable. 
If we end up needing to use the string "docker-ce.repo" multiple times (even if a REPO_FILE is defined beforehand) then we can create the extra variable.

seemethere: Historically we've set default values this way in this script, see lines 26-34

PR #74, opened 2018-06-07
WIP Switch over to supported file

seemethere:
This ultimately gets rid of the SUPPORT_MAP variable. It was tough to
manage and wasn't entirely representative of what distributions were
currently supported on which channels.

This commit makes it so that we call out to a text file that tells us
which architectures and distributions are currently supported for each
`VERSION` including the `latest` version.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -38,47 +38,6 @@ if [ -z "$REPO_FILE" ]; then
 	REPO_FILE="$DEFAULT_REPO_FILE"
 fi
 
-SUPPORT_MAP="
-x86_64-centos-7
-x86_64-fedora-26
-x86_64-fedora-27
-x86_64-fedora-28
-x86_64-debian-wheezy
-x86_64-debian-jessie
-x86_64-debian-stretch
-x86_64-debian-buster
-x86_64-ubuntu-trusty
-x86_64-ubuntu-xenial
-x86_64-ubuntu-bionic
-x86_64-ubuntu-artful
-s390x-ubuntu-xenial
-s390x-ubuntu-bionic
-s390x-ubuntu-artful
-ppc64le-ubuntu-xenial
-ppc64le-ubuntu-bionic
-ppc64le-ubuntu-artful
-aarch64-ubuntu-xenial
-aarch64-ubuntu-bionic
-aarch64-debian-jessie
-aarch64-debian-stretch
-aarch64-debian-buster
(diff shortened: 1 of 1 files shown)


seemethere:
*NOTE*: Tests will fail for this PR until some changes are made on the repository generation side.

seemethere:
Closing this, kind of outdated at this point.

install.sh
```diff
@@ -329,6 +307,10 @@ do_install() {
 			if [ -z "$dist_version" ] && [ -r /etc/lsb-release ]; then
 				dist_version="$(. /etc/lsb-release && echo "$DISTRIB_CODENAME")"
 			fi
+			# Install curl if it's not already installed since it's important for downloading
+			if ! command_exists curl; then
+				(set -ex; $sh_c "apt-get update && apt-get install -y curl")
```
delfer: fedora? centos?
`which curl || ( apt-get update && apt-get install -y curl) || yum install -y curl || zypper install -y curl || urpmi --auto curl || pacman --noconfirm -Sy curl || upgradepkg --install-new || apk --no-cache add curl`

nemchik: Good catch, but wow that solution feels messy. It looks like it would work, and as much as I love one liners when working in CLI, a script has a little more room to let code be code. Another thought is you've included things like pacman but this script doesn't support Arch for the moment. Maybe a more comprehensive PR to add full support for Arch would make more sense.

nemchik: Maybe something like this
```bash
if command_exists apt-get; then
    (set -ex; $sh_c "apt-get install -y curl")
elif command_exists dnf; then
    (set -ex; $sh_c "dnf install -y curl")
elif command_exists yum; then
    (set -ex; $sh_c "yum install -y curl")
fi
```

seemethere: You'll only reach this if you're running an old version of debian so I think the usage of `apt-get` is appropriate here

nemchik: So aside from an old version of debian all other supported distros already include curl?

seemethere: Oh wait nvm, I was confused this is only installation instructions for `ubuntu`,

screen shot 2019-01-23 at 9 33 18 am

nemchik: Wow... That's what happens when we don't scroll up...

Anyway would it be appropriate for there to be similar lines in the other distros? Or do they all include curl?

PR #79, opened 2018-10-16
Remove Ubuntu Artful since it is end of life.

roman-mueller:
Remove Ubuntu Artful since it is end of life.


seemethere:
@roman-mueller can you rebase this pull request? I merged in the latest shellcheck resolutions in #84

seemethere:
Thanks @roman-mueller!

Going to ignore the test results on this one, seems to only be failing on distributions we don't support anymore.

PR #84, opened 2018-11-08
Set to explictly disable all repos for RPM first

seemethere:
Makes it so that the repo that's enabled is only the channel that we
specify.

Was the case where it'd automatically enable stable without explicitly
asking if we want stable enabled.


seemethere:
Proceeding with the merge to get nightlies unblocked.

Known failures:
* ubuntu trusty
* debian jessie

PR #89, opened 2018-12-18
Made the suggested command usable

Nadav-Ruskin:
It always bothered me...


nemchik:
Shouldn't it be `$USER` ? (case sensitive)

seemethere:
yeah agree with @nemchik, closing due to inactivity though.

PR #91, opened 2019-01-16
TAR-238 Remove SUPPORT_MAP

corbin-coleman:
Just let the script run on whatever it's trying to be ran on.

Signed-off-by: corbin-coleman <corbin.coleman@docker.com>


nemchik:
There was a similar commit that was a little more involved https://github.com/docker/docker-install/pull/74

PR #95, opened 2019-02-14
add ee promo

seemethere:


install.sh
```diff
@@ -201,6 +201,36 @@ ee_notice() {
...
+			;;
+	esac
+	echo
+	echo "** Learn more at https://www.docker.com/products/docker-engine **"
+	echo
+	echo "ACTIVATE your own engine to Docker Engine - Enterprise using:"
+	echo
+	echo "  sudo docker engine activate"
+	echo
+	sleep 10
```
RobbKistler: Is the sleep necessary?

seemethere: Necessary for "visibility"

nemchik: The sleep is followed by an exit so it would be the last thing seen before returning to a prompt, and thus left on screen unless the user did some kind of reset or clear in sequence.

Issue #77, opened 2018-09-21
update the default channel in get.docker.com  


mataneja:
Currently, get.docker.com sets the default channel to edge

> DEFAULT_CHANNEL_VALUE="edge"

If we continue to plan to use this script, it would be useful to point this to Stable/Nightly instead

mataneja:
A lot of people continue to use this script for production deployments despite our Notice to the contrary. Perhaps, best to have 2 versions of this script or make the channel a dynamic argument on the CLI. This will let people use the script and select stable channel if deploying to prod.

MAGICCC:
Well you can use `curl -sSL https://get.docker.com/ | CHANNEL=stable sh`

PR #104, opened 2019-03-06
Use "stable" as default channel

thaJeztah:
Looks like "this value will automatically get changed" may
not be correct, as installing with this script enabled the
"edge" channel.

```
DRY_RUN=1 ./install.sh 
# Executing docker install script, commit: 40b1b76
...

yum install -y -q yum-utils
yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
yum-config-manager --disable docker-ce-*
yum-config-manager --enable docker-ce-edge
yum makecache
yum install -y -q docker-ce
```

fixes https://github.com/docker/docker-install/issues/77

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -19,11 +19,12 @@ set -e
 SCRIPT_COMMIT_SHA=UNKNOWN
 
 
-# This value will automatically get changed for:
-#   * edge
+# The channel to install from:
+#   * nightly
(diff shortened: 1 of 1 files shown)


thaJeztah:
ping @seemethere @andrewhsu PTAL

PR #106, opened 2019-03-15
[TAR-571] Remove the ads

seemethere:
This reverts commit 40b1b763998f478b3774d93763ad33eacc77def8, reversing
changes made to 26dda3d66987a707ec01f6fc2485bd0d379f7fee.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,6 +21,7 @@ def aarch64verifyTargets = [
   'aarch64-verify-install-ubuntu-xenial',
   'aarch64-verify-install-ubuntu-bionic',
   'aarch64-verify-install-centos-7',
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -21,6 +21,7 @@ def aarch64verifyTargets = [
   'aarch64-verify-install-ubuntu-xenial',
   'aarch64-verify-install-ubuntu-bionic',
   'aarch64-verify-install-centos-7',
+  'aarch64-verify-install-fedora-29',
```
StefanScherer: Do we reintroduce the fedora 29?

seemethere: No unfortunately I think that was the side-effect of adding this on top of PR #95, going to re-apply that particular patch

PR #110, opened 2019-04-09
TAR-666: Add version pinning for docker-ce-cli

seemethere:
Tested with the following versions:

```bash
#!/usr/bin/env bash

set -e

RUN="docker run --rm -it -v $(pwd):/v -w /v"

for distro in ubuntu:bionic debian:stretch centos:7 fedora:29; do
        for version in 18.09.2 18.06.1; do
                echo "${distro}, VERSION: ${version}"
                ${RUN} "${distro}" sh -c "VERSION=${version} sh install.sh && docker --version"
        done
done
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -367,12 +367,18 @@ do_install() {
 						echo
 						exit 1
 					fi
+					search_command="apt-cache madison 'docker-ce-cli' | grep '$pkg_pattern' | head -1 | cut -d' ' -f 4"
+					# Don't insert an = for cli_pkg_version, we'll just include it later
+					cli_pkg_version="$($sh_c "$search_command")"
 					pkg_version="=$pkg_version"
 				fi
 			fi
 			(
 				if ! is_dry_run; then
 					set -x
+				fi
+				if [ -n "$cli_pkg_version" ]; then
+					$sh_c "apt-get install -y -qq --no-install-recommends docker-ce-cli=$cli_pkg_version >/dev/null"
 				fi
 				$sh_c "apt-get install -y -qq --no-install-recommends docker-ce$pkg_version >/dev/null"
 			)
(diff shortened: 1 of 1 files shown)


seemethere:
@thaJeztah Since containerd.io doesn't necessarily correspond to a specific release of the engine/cli I'm going to not include it in the script. (it's too much business logic to maintain without adding extra env variables to tune)

thaJeztah:
Yes, not pinned to a version, more "make sure it's updated if we run the script"

seemethere:
@thaJeztah we could have the script run an upgrade for containerd.io if it's already installed but then again I think the overlap for people who already have containerd.io installed and would use this script to install docker-ce is probably pretty low.

install.sh
```diff
@@ -437,6 +446,10 @@ do_install() {
 				if ! is_dry_run; then
 					set -x
 				fi
+				# install the correct cli version first
+				if [ -n "$cli_pkg_version" ]; then
+					$sh_c "$pkg_manager install -y -q docker-ce-cli-$cli_pkg_version"
```
StefanScherer: @seemethere a = instead of - before $cli_pkg_version ?

seemethere: I'm just checking here if it's non-empty

thaJeztah: I think he meant;

```suggestion
					$sh_c "$pkg_manager install -y -q docker-ce-cli=$cli_pkg_version"
```

seemethere: Oh for apt it's an equals sign not a dash

thaJeztah: Yes, I missed it as well (good 👀 @StefanScherer 🤗)

seemethere: Wait sorry, lost track of where we were in the code, a `-` is the right character for `yum/dnf` install

thaJeztah: 🤦‍♂️ ah, you're right. Perhaps we should change `$pkg_manager` into `$yum` or `$yum_or_dnf` 😅

StefanScherer: Ah ok, didn't saw that from the phone that this is in the centos case. 🔍

Issue #112, opened 2019-04-10
rootless: move to 19.03 by default


tonistiigi:
We have 19.03 (test) tarballs of rootless extras now. The install script should start defaulting to that now, with a possibility to switch `CHANNEL` between nightly, test and stable.

https://download.docker.com/linux/static/test/x86_64/

@AkihiroSuda

zelahi:
@tonistiigi I took a small pass at this.  I hope it works or if there are any required changes, I would love your feedback!!

AkihiroSuda:
can we move this forward?

AkihiroSuda:
https://get.docker.com/rootless now defaults to 19.03.5. This issue is fixed.

StefanScherer:
Thanks for confirming.

PR #118, opened 2019-05-20
19.03 rootless install

zelahi:
**Description**
Based on issue # [112](https://github.com/docker/docker-install/issues/112), it looked like there was a need to install docker from multiple channels.  I took a pass at it.  Overall, it now provides three channels from which docker-ce can be installed: `test`, `stable`, and `nightly`.  

However it looks like the binaries need to be retagged as `docker.tgz` on: 
- https://download.docker.com/

### DIFF
--- a/rootless-install.sh
+++ b/rootless-install.sh
@@ -25,8 +25,6 @@ init_vars() {
 	if systemctl --user daemon-reload >/dev/null 2>&1; then
 		SYSTEMD=1
 	fi
-
-	## Add the Channel variable here
 }
 
 checks() {
@@ -171,9 +169,6 @@ For example:
 echo \"$(id -un):100000:65536\" >> /etc/subgid"
 		exit 1
 	fi
-
-
-	# Add checks here to Channel to make sure if was set.  If it wasn't set, then we need to default it to testing
 }
 
 start_docker() {
--- a/rootless-install.sh
+++ b/rootless-install.sh
@@ -306,15 +306,18 @@ do_install() {
 	tmp=$(mktemp -d)
 	trap "rm -rf $tmp" EXIT INT TERM
 
+	# Download docker binary based on the build the user requested
+	BASE_STATIC_RELEASE_URL="https://download.docker.com/linux/static"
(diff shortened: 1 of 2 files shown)


rootless-install.sh
```diff
@@ -304,10 +305,22 @@ do_install() {
...
-	# Alternatively could find latest nightly release from https://download.docker.com/linux/static/nightly/ .
-	# Later we can provide different channels.
-	STATIC_RELEASE_URL="https://master.dockerproject.org/linux/x86_64/docker.tgz"
-	STATIC_RELEASE_ROOTLESS_URL="https://master.dockerproject.org/linux/x86_64/docker-rootless-extras.tgz"
+
+	# Download docker binary based on the build the user requested
+	BASE_STATIC_RELEASE_URL="https://download.docker.com/linux/static"
+	MASTER_STATIC_RELEASE_URL="https://master.dockerproject.org/linux/x86_64"
+	if [ -z "$CHANNEL" ] || [ "$CHANNEL" = "test" ]; then
+					STATIC_RELEASE_URL="$BASE_STATIC_RELEASE_URL/test/x86_64/docker-19.03.0-beta4.tgz"
```
tonistiigi: these need to be automatically detected, not hardcoded

zelahi: detecting the channel or the `docker-19.03.0-beta4.tgz`?

tonistiigi: try to find the last build url from the list at `download.docker.com` for the specific channel

seemethere: Yeah we should probably also look into having just a latest `docker.tgz` in these as well.

zelahi: We could modify the jobs to upload a `docker.tgz`.  I'll try to find a link to the Jenkin jobs

AkihiroSuda:
Any update?

zelahi:
I think for now I'm going to close this PR.  We're looking at re-doing the infra right now.  So I think it would be better to open a new PR once the pathing and building for that is set in stone

PR #145, opened 2019-10-21
Skip docker.tgz download if docker is already installed

X0rg:
It allows user to use Docker version installed on his system.

### DIFF
--- a/rootless-install.sh
+++ b/rootless-install.sh
@@ -13,6 +13,9 @@
 # the script was uploaded (Should only be modified by upload job):
 SCRIPT_COMMIT_SHA=UNKNOWN
 
+# Set this variable to avoid to download docker.tgz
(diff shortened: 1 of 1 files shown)


rootless-install.sh
```diff
@@ -25,6 +25,11 @@ init_vars() {
 	if systemctl --user daemon-reload >/dev/null 2>&1; then
 		SYSTEMD=1
 	fi
+
+	USE_SYSTEM_DOCKER=
+	if command -v docker >/dev/null 2>&1; then
+		USE_SYSTEM_DOCKER=1
```
AkihiroSuda: This should not be set by default because of potential version mismatch

X0rg: Better in 1bb6e67?

PR #146, opened 2019-10-24
rootless: support channels (stable/nightly)

AkihiroSuda:
The default channel is set to "stable".

The latest stable version is currently hardcoded in the script.

Signed-off-by: Akihiro Suda <akihiro.suda.cz@hco.ntt.co.jp>


AkihiroSuda:
@zelahi @tonistiigi @tiborvass @seemethere PTAL

AkihiroSuda:
As a side-effect, this also supports non-amd64 archs

tonistiigi:
SGTM, but we should follow up with something that removes the constant version. Eg. `curl -sL https://download.docker.com/linux/static/stable/$(uname -m)  | grep -Po '(?<=>docker-rootless-extras-).*(?=\.tgz)' | sort -rV | head -n 1`

AkihiroSuda:
Parsing HTML with shell commands does not seem robust.
I think the server should have a symlink for the latest version.

AkihiroSuda:
@tonistiigi 
Can we get this merged?
We are highly likely to break master during migration to cgroup v2, so I want to switch the default to the stable channel.

zelahi:
Merged. Ty for the PR @AkihiroSuda

AkihiroSuda:
thanks, could you sync https://get.docker.com/rootless ?

PR #152, opened 2020-02-03
Use shellcheck stable image

StefanScherer:
This PR probably fixes the issue that we saw some months ago
```
shellcheck: src/ShellCheck/Analytics.hs:1404:10-27: No instance nor default method for class operation mappend
```
in https://ci.docker.com/public/blue/organizations/jenkins/docker-install/detail/PR-149/2/pipeline/
Use the stable shellcheck image instead.


StefanScherer:
The pending `continuous-integration/jenkins/pr-head` job is due to an outdated webhook. I've deactivated it just after this PR.

PR #153, opened 2020-02-13
rootless-install.sh: bump up to 19.03.6

AkihiroSuda:


StefanScherer:
Hm, we see this `continuous-integration/jenkins/pr-head Expected — Waiting for status to be report...` for the last couple of PR's over the last few months. I've checked the Jenkins config and compared it to similar pipelines we have, but cannot find the reason why Jenkins doesn't send a status.

PR #157, opened 2020-03-05
rootless-install.sh: bump up to 19.03.7

AkihiroSuda:


AkihiroSuda:
Can we somehow automate this?

I suggest adding the latest version string ("19.03.7") to  a text file like https://download.docker.com/linux/static/stable/x86_64/LATEST

AkihiroSuda:
@StefanScherer @thaJeztah @tonistiigi @tiborvass

thaJeztah:
> I suggest adding the latest version string ("19.03.7") to a text file like
> https://download.docker.com/linux/static/stable/x86_64/LATEST

Could be good to have; we may have to consider what works best (also if we decide to move containerd and possible the cli to a separate tar)

thaJeztah:
@tiborvass @silvin-lubecki ^^

silvin-lubecki:
The checks are still on `Waiting for status` 🤔cc @thaJeztah @StefanScherer

thaJeztah:
@silvin-lubecki yes, I think there's a bug on GitHub where, after you mark a check as "required", it will be _added_ to existing  PR's, but never actually run (or complete) 😞 

@AkihiroSuda could you do a force push to see if that helps?

StefanScherer:
I've kicked off https://ci-next.docker.com/teams-core/job/get.docker.com/job/upload-rootless-install/5/ 
Something we want to convert into a GH action soon :-)

thaJeztah:
Was thinking about that (didn't know how it got deployed). Would we need an option to make changes without them being deployed directly? (In case there's large changes for an upcoming release) guess we could use a feature branch for that or something 🤔

PR #160, opened 2020-03-06
Update GitHub Actions configuration

thaJeztah:
- GH Actions: set 5 minute timeout for the tests
  GitHub Actions have a default timeout of 6 hours, which is waaaay to high.
  This updates the timeout to 5 minutes (they currently run in just over a minutes, so for now this should be enough).
- GH Actions: remove 18.09 as it reached EOL

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,6 @@ jobs:
           - centos:7
         version:
           - 19.03
-          - 18.09
           - ""
 
     steps:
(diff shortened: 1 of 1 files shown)


thaJeztah:
I made the test steps "without version" _required_, and the other ones "optional" (versions may be updated once in a while, so this way we don't have to update the repository configuration each time)

thaJeztah:
It's green; merging

PR #175, opened 2020-05-31
rootless-install.sh: bump up to 19.03.10

AkihiroSuda:


AkihiroSuda:
Can we somehow automate this step?
I expect some file like `https://download.docker.com/linux/static/stable/x86_64/LATEST` to exist and contain the latest version number ("19.03.10").

StefanScherer:
Thank you @AkihiroSuda. It's a good idea to automate the few steps, but it's not our highest priority things. I'll add it to the list of things to do to improve the whole release pipeline.

PR #178, opened 2020-06-10
rootless-install.sh: use safer XRD on non-systemd hosts

AkihiroSuda:
`/tmp/docker-$(id -u)` is not a good candidate as `$XDG_RUNTIME_DIR`, because it might be already created by another user.

The new path is `$HOME/.docker/run`.

This commit does not affect systemd-based hosts.

### DIFF
--- a/rootless-install.sh
+++ b/rootless-install.sh
@@ -101,8 +101,8 @@ checks() {
 			>&2 echo "- or simply log back in as the desired unprivileged user (ssh works for remote machines)"
 			exit 1
 		fi
-		export XDG_RUNTIME_DIR="/tmp/docker-$(id -u)"
-		mkdir -p "$XDG_RUNTIME_DIR"
+		export XDG_RUNTIME_DIR="$HOME/.docker/run"
+		mkdir -p -m 700 "$XDG_RUNTIME_DIR"
 		XDG_RUNTIME_DIR_CREATED=1
 	fi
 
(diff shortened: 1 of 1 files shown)


rootless-install.sh
```diff
@@ -101,8 +101,8 @@ checks() {
 			>&2 echo "- or simply log back in as the desired unprivileged user (ssh works for remote machines)"
 			exit 1
 		fi
-		export XDG_RUNTIME_DIR="/tmp/docker-$(id -u)"
-		mkdir -p "$XDG_RUNTIME_DIR"
+		export XDG_RUNTIME_DIR="$HOME/.docker/xrd"
```
thaJeztah, line 104: As discussed on Slack (for a follow-up) I think we should find a way where not require `XDG_RUNTIME_DIR` to be set. This env-var is not "owned" by docker, so we should not set it. Instead, we could use a default path for the non-`XDG_RUNTIME_DIR` case (perhaps clean up the directory on daemon start)

rootless-install.sh
```diff
@@ -307,6 +307,7 @@ print_instructions() {
 	echo "# Make sure the following environment variables are set (or add them to ~/.bashrc):\n"
 
 	if [ -n "$XDG_RUNTIME_DIR_CREATED" ]; then
+		echo "# WARN: systemd not found. You have to remove XDG_RUNTIME_DIR manually on every logout."
```
thaJeztah, line 310: Should we print the actual path (value of `XDG_RUNTIME_DIR`) instead (or "as well")?

AkihiroSuda: one line below :)

rootless-install.sh
AkihiroSuda: Do we want that feature in 20.0X? or in 20.1X? (21.0X?)

thaJeztah: If we have it ready before 20.0x, I think that'd be good to have

AkihiroSuda:
Addressed s/xrd/run/ (https://github.com/moby/moby/pull/41098#discussion_r440845147)

thaJeztah:
@silvin-lubecki @StefanScherer PTAL

PR #181, opened 2020-06-26
rootless: call dockerd-rootless-setuptool.sh for $CHANNEL=nightly

AkihiroSuda:
`dockerd-rootless-setuptool.sh` is the new standard way to set up the systemd unit file.
https://github.com/moby/moby/blob/master/contrib/dockerd-rootless-setuptool.sh

The usage of `rootless-install.sh` stays same.

### DIFF
--- a/rootless-install.sh
+++ b/rootless-install.sh
@@ -55,6 +55,11 @@ init_vars() {
 	if systemctl --user daemon-reload >/dev/null 2>&1; then
 		SYSTEMD=1
 	fi
+
+	V1903=
+	if [ "$CHANNEL" = "stable" ] && ( echo "$STABLE_LATEST" | grep -q "^19\.03" ); then
+		V1903=1
+	fi
 }
 
 checks() {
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
@thaJeztah @tianon PTAL

rootless-install.sh
```diff
@@ -327,7 +338,21 @@ print_instructions() {
...
+	if [ -n "$FORCE_ROOTLESS_INSTALL" ]; then
+		args="$args --force"
+	fi
+	if [ -n "$SKIP_IPTABLES" ]; then
+		args="$args --skip-iptables"
+	fi
+	set -x
+	(
+		"$BIN/dockerd-rootless-setuptool.sh" install $args
+	)
```
tianon, line 355: Instead of this extra `args` variable, I think we should just use `set`, especially since it will properly preserve any whitespace, which isn't relevant right now but could be in the future (also, the `set -x` should be inside the subshell or it will leak and the subshell won't do us any good :sweat_smile:):

```suggestion
	if [ -n "$FORCE_ROOTLESS_INSTALL" ]; then
		set -- "$@" --force
	fi
	if [ -n "$SKIP_IPTABLES" ]; then
		set -- "$@" --skip-iptables
	fi
	(
		set -x
		"$BIN/dockerd-rootless-setuptool.sh" install "$@"
	)
```

AkihiroSuda: thanks, updated

AkihiroSuda:
ping @thaJeztah @tiborvass  @tonistiigi

AkihiroSuda:
@StefanScherer PTAL?

StefanScherer:
@thaJeztah Could you also take a look. Is it future proof for v20.xx? Or good enough for now?

AkihiroSuda:
Should be good enough for now

PR #182, opened 2020-06-30
Retrieve latest docker release

StefanScherer:
Instead of adjusting the version number after each release, what about retrieving the latest release from GitHub automatically?

@AkihiroSuda WDYT?

### DIFF
--- a/rootless-install.sh
+++ b/rootless-install.sh
@@ -27,8 +27,13 @@ DEFAULT_CHANNEL_VALUE="stable"
 if [ -z "$CHANNEL" ]; then
 	CHANNEL=$DEFAULT_CHANNEL_VALUE
 fi
-# The latest release is currently hard-coded.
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
This will be broken when we have 19.03.x backport releases after 20.03.x?

StefanScherer:
Bummer, yes thanks.
That would probably happen as well when we add the upload into our release CI.
OK, let's think about a better solution. I'm closing this for now.

AkihiroSuda:
We could do this if we can depend on jq, but I don't think we want to depend on jq
```console
$ curl -s https://api.github.com/repos/docker/docker-ce/releases | jq -r '.[].name' | sed -e s/^v//g | grep ^19.03 | head -n 1
19.03.12
```

StefanScherer:
Thanks, yes we don't want to depend on `jq` here.

PR #192, opened 2020-09-22
fix typo in rootless-install.sh

oToToT:
I think that the \n inside `echo "# Make sure the following environment variables are set (or add them to ~/.bashrc):\n"` is unnecessary.  
Maybe we should remove it.


AkihiroSuda (approved):

StefanScherer (approved):
Thank you.

Issue #183, opened 2020-07-29
why to sleep for 20 seconds?


vladyslavstartsev:
I see this line https://github.com/docker/docker-install/blob/2e839c9ac3e469baedcf74069d0e95200957d332/install.sh#L265
 and also others. I don't really understand why should I sleep for 20 sec?

thaJeztah:
Because otherwise the user wouldn't be able to act on the warnings that are printed; https://github.com/docker/docker-install/blob/2e839c9ac3e469baedcf74069d0e95200957d332/install.sh#L237-L258

vladyslavstartsev:
Oh, I was using this script in my dotfiles, and thought that I wait too long

nemchik:
I use the script in a similar way and actually think it would be nice to be able to bypass the sleep by setting an environment variable or passing a flag or something.

thaJeztah:
What's the reason you're getting the error? If you're using the script to update to a newer version, using the regular way to update packages (`apt-get update && apt-get install -y docker-ce docker-ce-cli containerd.io`, or the equivalent yum/dnf commands) might be better

nemchik:
The sleep happens when this condition is met https://github.com/docker/docker-install/blob/2e839c9ac3e469baedcf74069d0e95200957d332/install.sh#L221 even though one of the potential scenarios already accounted for displays a message informing the user they can safely ignore the message https://github.com/docker/docker-install/blob/2e839c9ac3e469baedcf74069d0e95200957d332/install.sh#L257 in which scenario I would ideally want the option to also skip the sleep when calling this script as part of another script's installation procedure. 20 seconds isn't hurting anyone, but if the option to skip it is being considered I can confirm it would be useful and would be used.

Also just to clarify, what I'm asking is along the lines of setting an environment variable prior to running the script that would be something like `SKIP_DOCKER_CE_INSTALL_SLEEP_WARNING=true` or something that's very intentional. I would also consider some kind of flag when running like `bash docker-install.sh --skip-sleep-when-docker-already-exists=true` (maybe not that long as a flag, but point being it wouldn't be something passed on accident like `-s`).

thaJeztah:
> Oh, I was using this script in my dotfiles,

I'm still curious when the script is called / used in your use-case to get a better understanding; Is it to keep your docker installation up-to-date? How do you keep other packages up to date (and wouldn't docker be included in the same process?)

>  informing the user they can safely ignore the message

[...]

PR #195, opened 2020-10-14
Explicitly install both containerd and Docker CLI to handle upgrades

tiborvass:
Signed-off-by: Tibor Vass <tibor@docker.com>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -391,8 +391,7 @@ do_install() {
 						exit 1
 					fi
 					search_command="apt-cache madison 'docker-ce-cli' | grep '$pkg_pattern' | head -1 | awk '{\$1=\$1};1' | cut -d' ' -f 3"
-					# Don't insert an = for cli_pkg_version, we'll just include it later
-					cli_pkg_version="$($sh_c "$search_command")"
+					cli_pkg_version="=$($sh_c "$search_command")"
 					pkg_version="=$pkg_version"
 				fi
 			fi
@@ -400,10 +399,7 @@ do_install() {
 				if ! is_dry_run; then
 					set -x
 				fi
-				if [ -n "$cli_pkg_version" ]; then
-					$sh_c "apt-get install -y -qq --no-install-recommends docker-ce-cli=$cli_pkg_version >/dev/null"
-				fi
-				$sh_c "apt-get install -y -qq --no-install-recommends docker-ce$pkg_version >/dev/null"
+				$sh_c "apt-get install -y -qq --no-install-recommends docker-ce-cli$cli_pkg_version containerd.io docker-ce$pkg_version >/dev/null"
 			)
 			echo_docker_as_nonroot
 			exit 0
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -460,7 +456,7 @@ do_install() {
 					fi
 					search_command="$pkg_manager list --showduplicates 'docker-ce-cli' | grep '$pkg_pattern' | tail -1 | awk '{print \$2}'"
 					# It's okay for cli_pkg_version to be blank, since older versions don't support a cli package
-					cli_pkg_version="$($sh_c "$search_command" | cut -d':' -f 2)"
+					cli_pkg_version="-$($sh_c "$search_command" | cut -d':' -f 2)"
```
tiborvass, line 459: this probably breaks installing older versions

thaJeztah (approved):
SGTM, but not at my computer to test.

thaJeztah:
@tianon ptal if you have time <3

thaJeztah:
After this, we may be able to address https://github.com/docker/docker-install/issues/183 as well

tianon:
I'm not really in-tune with how folks are using this script, but I know installing old versions used to be really common.

I agree that updating `containerd` too is good, but I worry that it might be a little risky, for example when we get 1.4 in the repo, will older Docker still be compatible?  Should we update `Depends` on Docker to set an upper bound? (Not sure how well APT handles that, actually)

seemethere:
Why is this necessary? If users wanted to upgrade their containerd.io version then they'd run a package upgrade 

Also if a new install of docker-ce requires a newer version of containerd then it should upgrade through the package manager. 

Not entirely sure I see the benefit here.

thaJeztah:
> for example when we get 1.4 in the repo, will older Docker still be compatible? Should we update Depends on Docker to set an upper bound? (Not sure how well APT handles that, actually)

We discussed that recently, and @cpuguy83 tried that, but unfortunately `apt` (at least) doesn't support that. Well, it supports restricting what's compatible (`Depends containerd.io < 1.4`), but when installing, will always install "latest" version of `containerd.io`, and after that fail (because then it detects the version doesn't match the requirement).

Not sure if there's alternatives to that approach. On the other hand, we discussed if we should forcibly _block_ users from installing newer containerd versions when installing older versions of docker; given that "technically", newer versions should be compatible, we thought it was best to leave "holding back" newer versions as an exercise the the user.

> I'm not really in-tune with how folks are using this script,

and

> Why is this necessary? If users wanted to upgrade their containerd.io version then they'd run a package upgrade

Yes, that _should_ be the normal way to upgrade docker, containerd (or any package). However, in practice I frequently encounter users using it to upgrade docker as well (or even use it in their dotfiles? 🤷‍♂️ https://github.com/docker/docker-install/issues/183#issuecomment-665661341)

The problem with that situation is that (even though a warning is printed and the script is halted for 20 seconds), it "seems" to work (docker is updated), however, users end up with a _partial_ update only (e.g. "engine" updated, but cli not, or containerd not being updated).


> but I know installing old versions used to be really common.

Perhaps we need the approach that Rancher took, and have a script for each version ? Not a huge fan from a maintenance perspective thought https://github.com/rancher/install-docker

tianon:
I mean, I'm still a fan of deprecating the script entirely, which makes the maintenance burden go away completely. :trollface: :heart:

(Really, pushes the burden back onto the docs, which is something needed anyhow.)

tiborvass:
Agree for containerd, but not for docker cli, the CLI is more docker than engine these days. I'll update to not install containerd.io explicitly.

PR #197, opened 2020-10-22
make sure certificates are up-to-date

thaJeztah:
Found a TODO about this in one of our internal test-scripts to include this step in the get.docker.com install script.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -369,6 +369,8 @@ do_install() {
 				fi
 				$sh_c 'apt-get update -qq >/dev/null'
 				$sh_c "DEBIAN_FRONTEND=noninteractive apt-get install -y -qq $pre_reqs >/dev/null"
+				# Make sure ca-certificates are up-to-date
+				$sh_c "update-ca-certificates -f"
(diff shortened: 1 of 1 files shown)


thaJeztah:
@tianon @tiborvass PTAL 😅

tianon:
Honestly, I'm a little confused what problem this is solving?  If there's a package that gets installed that needs this to be run, it'll usually run it automatically via `postinst`. :confused:

Is there an issue/report somewhere with more details on what this is fixing?

thaJeztah:
@tianon I dug into history for that change, and it was added to fix issues in our builds for the armhf builds (either the host machines or in the balena images, I don't recall). Apparently those images at some points had (curl and) ca-certificates already installed, but the certificates were outdated. Then running curl (?) failed because the certificates were outdated.

I must admit that I don't recall the exact details, and the Jenkins logs for those failures are long gone. I'm ok with closing, just wanted to open this one for discussion to see if it would make sense to include (perhaps it would be similar scenarios where users run into the same problem?)

tianon:
Yeah, IMO this sounds like a bug in either the base image / environment or one of the packages being installed. :confused:

thaJeztah:
Let's close this one for now

PR #220, opened 2021-03-13
Use add-apt-repository if available.

dHannasch:
add-apt-repository is capable of seamlessly handling proxy settings.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -384,7 +384,7 @@ do_install() {
 				$sh_c 'apt-get update -qq >/dev/null'
 				$sh_c "DEBIAN_FRONTEND=noninteractive apt-get install -y -qq $pre_reqs >/dev/null"
 				$sh_c "curl -fsSL \"$DOWNLOAD_URL/linux/$lsb_dist/gpg\" | apt-key add -qq - >/dev/null"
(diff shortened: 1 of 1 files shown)


seemethere:
iirc the reason we had originally decided against it was that `add-apt-repository` was bugged for `arm` variations of `ubuntu` and `debian`.

seemethere:
Related issue: https://github.com/moby/moby/issues/31405

thaJeztah:
(Going through some old PRs); yes, I think this was A) because of the incompatibility mentioned, B) to reduce the number of dependencies needed to install (just for adding a line to a file), and C) manually creating the file gives a consistent filename for users to look for the repository config (it will always be in `/etc/apt/sources.list.d/docker.list`), which helps in some situations (e.g. when upgrading distros etc).

Let me close this PR, but thanks for contributing!

dHannasch:
Actually, taking another look at this, I think we can totally make this work manually.

dHannasch:
I don't understand the question (is there a difference?), but regardless, looks like the automated install totally works without this change, including behind a proxy. https://gitlab.com/shell-bootstrap-scripts/register-gitlab-runner/-/blob/master/install-docker.sh#L69
I thought that I'd had to use this branch to make it work, but I must have misremembered.

PR #223, opened 2021-04-09
Fixed the actual download address of the packages.

### LINKED ISSUE #213: 国内用户使用get-docker.sh安装docker-ce超时怎么办？
解决办法：
        编辑get-docker.sh文件，大约在437行附近增加一行:
`sed -i 's/download.docker.com/mirrors.aliyun.com\/docker-ce/g' /etc/yum.repos.d/docker-ce.repo`

# 增加示例如下:
`432                         (
433                                 if ! is_dry_run; then
434                                         set -x
435                                 fi
436                                 $sh_c "$pkg_manager install -y -q $pre_reqs"
437                                 $sh_c "$config_manager --add-repo $yum_repo"
438                                 **sed -i 's/download.docker.com/mirrors.aliyun.com\/docker-ce/g' /etc/yum.repos.d/docker-ce.repo**
439 
440                                 if [ "$CHANNEL" != "stable" ]; then
441                                         $sh_c "$config_manager $disable_channel_flag docker-ce-*"
442                                         $sh_c "$config_manager $enable_channel_flag docker-ce-$CHANNEL"
443                                 fi
444                                 $sh_c "$pkg_manager makecache"
445                         )`

tao12345666333:
fix #213

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -454,7 +454,16 @@ do_install() {
 				fi
 				$sh_c "$pkg_manager install -y -q $pre_reqs"
 				$sh_c "$config_manager --add-repo $yum_repo"
-
+                if [ -n "$mirror" ]; then
+                    case "$mirror" in
+                        Aliyun)
+                            sed -i "s#download.docker.com#mirrors.aliyun.com/docker-ce#g" /etc/yum.repos.d/docker-ce.repo
+                            ;;
+                        AzureChinaCloud)
+                            sed -i "s#download.docker.com#mirror.azure.cn/docker-ce#g" /etc/yum.repos.d/docker-ce.repo
+                            ;;
+                    esac
+                fi
 				if [ "$CHANNEL" != "stable" ]; then
 					$sh_c "$config_manager $disable_channel_flag docker-ce-*"
 					$sh_c "$config_manager $enable_channel_flag docker-ce-$CHANNEL"


tao12345666333:
@thaJeztah PATL, thanks! 

Before the modification, only the mirror was used to download the yum repo file, but the download of the rpm package was still through download.docker.com.

install.sh
```diff
@@ -454,7 +454,16 @@ do_install() {
...
+                if [ -n "$mirror" ]; then
+                    case "$mirror" in
+                        Aliyun)
+                            sed -i "s#download.docker.com#mirrors.aliyun.com/docker-ce#g" /etc/yum.repos.d/docker-ce.repo
+                            ;;
+                        AzureChinaCloud)
+                            sed -i "s#download.docker.com#mirror.azure.cn/docker-ce#g" /etc/yum.repos.d/docker-ce.repo
+                            ;;
+                    esac
+                fi
```
thaJeztah, line 466: Looks like the indentation is using spaces instead of tabs here.

Also wondering if we should repeat the `case` here, or if we should (e.g.) add a `RPM_REPO_URL` variable to the existing code higher up; https://github.com/docker/docker-install/blob/6580690119f66d3e31b8bc84e7ac9f9e3d68559b/install.sh#L63-L70 (or even re-use `DOWNLOAD_URL` to always update the `docker-ce.repo` URL?

And if that variable is set to do the replace.

thaJeztah: More in general though, I'm a bit unsure if we should keep these mirror options; are they still useful? I recall that support for our CDN's has improved behind the GFW, so perhaps it's no longer needed for "most" situations.

Looking at the mirrors, is the aliyun mirror still operational?

- https://mirrors.aliyun.com/docker-ce redirects to http://pub.mirrors.aliyun.com/docker-ce/ (does not resolve for me)
- https://mirror.azure.cn/docker-ce (looks to be working still)

There's also the rootless-install script; which has the URLs hard-coded; https://github.com/docker/docker-install/blob/6580690119f66d3e31b8bc84e7ac9f9e3d68559b/rootless-install.sh

Which adds a bit to my concern if adding extra complexity is worth it, or if we should remove it altogether.

We already have the `DOWNLOAD_URL` variable (and/or `RPM_REPO_URL`); if we added the replace for `docker-ce.repo`, we could remove the switch for these specific mirrors


Will have to check with others what's the desired approach

thaJeztah (changes requested):

wojiushixiaobai:
Thank you very much. At present, CDN can work normally in some unrestricted areas, http://pub.mirrors.aliyun.com/docker-ce/ Services are no longer available

PR #228, opened 2021-04-29
Handle CentOS forks

nunix:
Since CentOS 8 going into a rolling release mode with CentOS Stream, new forks have appeared for keeping the "stable release".

The first one to be available is Almalinux and the current script does not support it.

This PR addresses this gap, and as I don't yet the name/IDs of the future forks, I decided to go with a `case` statement, as it will be easier to manage in the long run.

By adding the `case` statement in the `get_distribution()` function, this avoids any impact on the creation of the distribution URL used in the script.

Finally, this fixes the issue in the Moby repo: https://github.com/moby/moby/issues/42337

Signed-off-by: nunix corsair@hey.com

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -120,6 +120,13 @@ get_distribution() {
 	if [ -r /etc/os-release ]; then
 		lsb_dist="$(. /etc/os-release && echo "$ID")"
 	fi
+	# Updates the "lsb_dist" variable for the CentOS forks
+	# new IDs will be updated in the future.
+	case "$lsb_dist" in
+		almalinux)
+			lsb_dist="centos"
+			;;
+	esac
 	# Returning an empty string here should be alright since the
 	# case statements don't act unless you provide an actual value
 	echo "$lsb_dist"


thaJeztah:
I'm a bit cautious on adding support for forks; we did have code for Debian derivatives in the past, and it was hell; things would often break (either due to "matching" not picking the right variant, or because the forks we _just_ slightly different), adding forks to the script gives the suggestion that they're _supported_, but we can't give any guarantees.
(also see https://docs.docker.com/engine/install/#other-linux-distributions)


/cc @chris-crone

chris-crone:
Yes, I am a bit worried about maintaining this over time as well. Could we have a manual override or something to make it explicit that you're opting into installing packages on an unsupported distro?

e.g.: `curl https://get.docker.com | DISTRO=centos sh`

thaJeztah:
Manual override _could_ possibly work.

Quick solution would be to change the error message, and add a link to that "other distros" section, so that users can follow the manual instructions for the distro that's the closest match

nunix:
thanks for the feedback and indeed, I went ahead as I saw there was already some fork handling.
Another possible way, would be to check the ID_LIKE as we will have "debian" for Ubuntu for example, and "rhel centos fedora" in the case of Almalinux.

Might be more in-line with the concept of checking the `/etc/os-release` file?

jochenw:
I understand, that this PR assumes the presence of /etc/redhat-release on CentOS forks? Confirming, that this should work on AlmaLinux, thus supersedes https://github.com/moby/moby/issues/42337.

nunix:
hi @jochenw , indeed, I changed the detection to be more "Red Hat standard" oriented.
I tested it with AlmaLinux and lately with Rocky Linux too and both went fine. But please feel free to test it on your own.

Please note also the valid concerns from the Docker team about forks.

jochenw:
@nunix This is, most likely, not a good place to ask. On the other hand, as we are already discussing CentOS, and compatibles:

Did you address the necessity to do a "dnf -y remove runc"?

nunix:
No, I didn't change any follow-up process that the scripts runs. Just the detection of forks and change the OS ID to `centos`.
If the script was doing it already for CentOS, then it will continue doing it for the forks.

jochenw:
Okay, I have created #230.

markd69:
Any news regarding this? It's hurting our deployments! 

No reason why this shouldn't be merged.

austin-millan:
Any way to get this this script working with AlmaLinux, a RHEL fork?

```bash
$ cat /etc/redhat-release 
AlmaLinux release 9.0 (Emerald Puma)
$ curl -fsSL https://raw.githubusercontent.com/docker/docker-install/d4716364808e26093a8fde88ebc94cc99b5cef3f/install.sh -o install-docker.sh
$ sudo sh ./install-docker.sh 

ERROR: Unsupported distribution 'forked distribution, switching to centos id
centos
```

markd69:
> Any way to get this this script working with AlmaLinux, a RHEL fork?
> 
> ```shell
> $ cat /etc/redhat-release 
> AlmaLinux release 9.0 (Emerald Puma)
> $ curl -fsSL https://raw.githubusercontent.com/docker/docker-install/d4716364808e26093a8fde88ebc94cc99b5cef3f/install.sh -o install-docker.sh
> $ sudo sh ./install-docker.sh 
> # Executing docker install script, commit: 
> 
> ERROR: Unsupported distribution 'forked distribution, switching to centos id
> centos
> ```

Check the commits tab on this pull request and make the changes to your own version of the script. Docker doesn’t appear to want to merge it, or add centos 9 support.

AkihiroSuda:
An alternative proposal:
- https://github.com/docker/docker-ce-packaging/issues/933

The support term of CentOS Stream is much shorter than Rocky and Alma, so it is probably inevitable to create RPMs for Rocky and Alma without depending on CentOS Stream.

Bart-De-Pauw:
Any plans on also including Oracle Linux Server?

```
$ cat /etc/os-release | head -3
NAME="Oracle Linux Server"
VERSION="9.4"
ID="ol"
```

Manual installation of docker on this release works perfect with the CentOS distribution.

PR #234, opened 2021-06-03
rootless-install.sh: use command -v instead of which

AkihiroSuda:
The which command is often missing.

ref: https://github.com/moby/moby/pull/42315


thaJeztah (approved):
LGTM, thanks!

PR #235, opened 2021-06-03
deb: explicitly install docker-ce-cli, scan-plugin, and docker-ce

thaJeztah:
internal ticket: [POS-27]

supersedes / replaces https://github.com/docker/docker-install/pull/195 (except for the "containerd.io" part)
closes https://github.com/docker/docker-install/pull/195

The scripts is using "--no-install-recommends" to provide a more minimal install, and to prevent installing "latest" version of (e.g.) the engine when installing the CLI.

This patch:

- adds some utilities to compare versions
- skips CLI-version discovery on older versions (before 18.09)  which did not have separate packages for the CLI
- on versions 20.10 and up (or if no version was specified), also install the scan cli plugin (support for cli plugins was added   in docker 19.03, but not installing on that version by default)) 

[POS-27]: https://docker.atlassian.net/browse/POS-27

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -21,6 +21,8 @@ set -e
 # the script was uploaded (Should only be modified by upload job):
 SCRIPT_COMMIT_SHA="${LOAD_SCRIPT_COMMIT_SHA}"
 
+# strip "v" prefix if present
+VERSION="${VERSION#v}"
 
 # The channel to install from:
 #   * nightly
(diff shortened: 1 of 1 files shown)


thaJeztah:
@chris-crone @glours PTAL

chris-crone (commented):
I ran a quick test with an Ubuntu container and it worked as expected. I didn't have a chance to test the version code path though

thaJeztah:
rebased, after https://github.com/docker/docker-install/pull/236 was merged

chris-crone (approved):
Tested the version path a bit and LGTM!

install.sh
```diff
@@ -69,16 +71,57 @@ case "$mirror" in
...
-has_rootless_extras="1"
-if echo "$VERSION" | grep -q '^1'; then
-	has_rootless_extras=
-fi
-
 command_exists() {
 	command -v "$@" > /dev/null 2>&1
 }
 
+# is_at_least_version checks if the version specified in $VERSION is at least
```
thaJeztah, line 78: oh! forgot to update this comment (renamed the function at some point); let me fix

thaJeztah: done 👍

glours (approved):
LGTM

PR #236, opened 2021-06-03
remove warning about pre-1.10 update

thaJeztah:
When upgrading docker from versions older than v1.10, the image store must be migrated for the content-addressable store.

Docker 1.10 was released more than 5 years ago, so people running versions older than that, probably have more urgent things to be concerned about, so removing this warning.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -221,33 +221,10 @@ check_forked() {
 	fi
 }
 
-semverParse() {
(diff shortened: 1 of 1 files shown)


thaJeztah:
@chris-crone @glours PTAL

chris-crone (approved):
LGTM and agree!

PR #240, opened 2021-07-13
Add mirror provided by Tencent

ldsink:
Add mirror provided by Tencent if mirror option is present.

usage:
> sudo ./install.sh --mirror Tencent


wizardbyron (approved):

ldsink:
@thaJeztah Could you please merge this PR?

thaJeztah (changes requested):

PR #241, opened 2021-07-17
Add support for debian bullseye and deprecate apt-key

YadominJinta:


thaJeztah:
Arf, my bad; pushed to the wrong branch; carried in https://github.com/docker/docker-install/pull/243

PR #250, opened 2021-08-16
Added a generic fix for enabling SLES repos

anujajakhade:
Added a generic fix for enabling SLES repos

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -536,8 +536,10 @@ do_install() {
 				echo "Packages for SLES are currently only available for s390x"
 				exit 1
 			fi
+
+			sles_version="${dist_version##*.}"
 			sles_repo="$DOWNLOAD_URL/linux/$lsb_dist/$REPO_FILE"
-			opensuse_repo="https://download.opensuse.org/repositories/security:SELinux/SLE_15_SP2/security:SELinux.repo"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -536,8 +536,11 @@ do_install() {
 				echo "Packages for SLES are currently only available for s390x"
 				exit 1
 			fi
+
+			zypper install -y lsb-release 
+			sles_version=$(lsb_release -r |  rev | cut -c1)
```
thaJeztah, line 541: If we need `lsb-release`, it should be added to the `pre_reqs` instead, but I think we can get this information without installing `lab-release`; 

We already should have `$dist_version`  available, which is set here; https://github.com/docker/docker-install/blob/0e685c6ac0bddd7b2ba7bcaaeb519746ad249a29/install.sh#L351-L353

In a SLES container, that prints;

```bash
dist_version="$(. /etc/os-release && echo "$VERSION_ID")"
echo $dist_version
15.3
```

Taking the suffix (after `.`);

```bash
sles_version="${dist_version##*.}"
echo $sles_version
3
```

So I think we can change this to:

```suggestion
			sles_version="${dist_version##*.}"
```

thaJeztah (changes requested):
Left a suggestion

thaJeztah (approved):
LGTM, thanks!

Issue #271, opened 2022-02-18
docker install doesn't work in debian 11


RodrigoTomeES:
Hi,

I have seen a lot of issues related with this #186 #187 #262 #265 but it is still not fixed.

![imagen](https://user-images.githubusercontent.com/13077343/154751579-5a72a146-d154-410a-9c9d-a4e0f152ce24.png)

My system is a Raspberry Pi 4 with 4GB of RAM

Linux raspberrypi 5.10.92-v8+ #1514 SMP PREEMPT Mon Jan 17 17:39:38 GMT 2022 aarch64 GNU/Linux

Please could you fix it?

Thanks!

Issue #270, opened 2022-01-27
Docker cannot be downloaded due to debian version 11


vedimuthan:
Hi,

We have an issue downloading docker through https://test.docker.com/ script. Please see the Dockerfile content below. When i do "cat /etc/debian_version", it shows 11.2. But if you look at the actual script, there is no logic for version 11. I could see it is only checking for version 8, 9 and 10. we are getting "**The repository 'https://download.docker.com/linux/debian 11 Release' does not have a Release file**" error. Can you please fix this issue?

```docker
FROM python:3.7

RUN apt-get update -y \
 && apt-get install -y \
        gnupg \
        gnupg2 \
 && apt-get clean

RUN pip install pip==20.2.3

# install docker
RUN cat /etc/debian_version
RUN curl -fsSL https://test.docker.com/ | sh
```

![image](https://user-images.githubusercontent.com/53485013/151433196-24dcca79-5794-4123-9eca-10825b8189b5.png)
![image](https://user-images.githubusercontent.com/53485013/151434643-4bf726fe-19cb-495f-9dcf-9f49a0023117.png)
![image](https://user-images.githubusercontent.com/53485013/151433739-6ba9a62e-03d5-4337-a616-093b50009c77.png)

thaJeztah:
This should be fixed through https://github.com/docker/docker-install/pull/243

Issue #262, opened 2021-11-14
failed to install on a rpi 4b with the latest rasperry pi os bullseye


pvklink:
The convenient script does not work with the latest rpi OS (full version) bullseye

PR #281, opened 2022-04-08
Add Debian Codename resolving

iamjla:
As mentioned in some issues #262 #270 #271 
The debian codename resolving is somewhat broken.
This is a stable fix which works down to debian squeeze.

It may also work for ubuntu but this is only the debian integration.


install.sh
```diff
@@ -330,21 +316,7 @@ do_install() {
...
-					dist_version="buster"
-				;;
-				9)
-					dist_version="stretch"
-				;;
-				8)
-					dist_version="jessie"
-				;;
-			esac
+			dist_version="$(sed -n 's/^VERSION_CODENAME=//p' < /etc/os-release )"
```
thaJeztah, line 319: This could take the same approach as `get_distribution()`; https://github.com/docker/docker-install/blob/0225270d2de23f66c4f551d97e74e9f831adfdc8/install.sh#L163-L172

```bash
if [ -r /etc/os-release ]; then
	lsb_dist="$(. /etc/os-release && echo "$VERSION_CODENAME")"
fi
```

However, I'm not sure how reliable codename is here, taking `sid` into account (on Debian); not sure if `raspbian` also uses `sid`, but if so, then it's likely that the codename doesn't always match the version.

iamjla: That is correct. Sid does not set the VERSION_CODENAME in /etc/os-release. 
raspbian uses the codename of the next release "bookworm" or "testing" instead of sid. 
However these versions are not present at download.docker.com which would let the script fail at that point. (or is there a testing/unstable repo?)


I would mimic the "lsb_dist=" style with the debian dist_version. lsb_dist would be for getting the os type while dist_version would get the debian/raspbian codename as ubuntu still behaves drifferent.

```suggestion
			dist_version="$(. /etc/os-release && echo "$VERSION_CODENAME")"
```

Would that be alright?

PR #298, opened 2022-06-07
install docker-buildx-plugin on version 20.10 as well, to allow upgrading

thaJeztah:
### install docker-buildx-plugin on version 20.10 as well, to allow upgrading

The `docker-buildx-plugin` package can replace the version of buildx that's
bundled in the `docker-ce-cli` packages. This allows for buildx to be updated
to the current version, even if it was originally installed from the bundled
version.

Some testing shows that this works well, and that upgrading the `docker-ce-cli`
package won't revert the change, and continues using the buildx version from the
separate package.

Here's some steps to illustrate that scenario:

```console
# remove any installed docker packages
apt-get purge -y docker-ce docker-ce-cli containerd.io docker-compose-plugin docker-scan-plugin docker-buildx-plugin docker-ce-rootless-extras

# stop the docker.socket service (it blocks a clean reinstall after an uninstall)
systemctl stop docker.socket

# install docker 20.10.14 (from the test channel)
curl -fsSL https://test.docker.com | VERSION=20.10.14 sh

# verify that the installed version of buildx is the bundled (`-docker`) version:
docker buildx version
github.com/docker/buildx v0.8.1-docker 5fac64c2c49dae1320f2b51f1a899ca451935554

# install the separate `docker-buildx-plugin` package
apt-get install -y docker-buildx-plugin

# verify that it replaced the bundled version
docker buildx version
github.com/docker/buildx v0.8.2

# upgrade to docker 20.10.17
curl -fsSL https://test.docker.com | VERSION=20.10.17 sh

# verify that updating didn't downgrade the buildx plugin, or revert it to the "included" buildx version:
docker buildx version
github.com/docker/buildx v0.8.2
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -538,10 +538,7 @@ do_install() {
 						pkgs="$pkgs docker-ce-cli${cli_pkg_version%=} containerd.io"
 				fi
 				if version_gte "20.10"; then
-						pkgs="$pkgs docker-compose-plugin docker-ce-rootless-extras$pkg_version"
-				fi
-				if version_gte "23.0"; then
-						pkgs="$pkgs docker-buildx-plugin"
+						pkgs="$pkgs docker-buildx-plugin docker-compose-plugin docker-ce-rootless-extras$pkg_version"
 				fi
 				if ! is_dry_run; then
 					set -x
@@ -628,10 +625,7 @@ do_install() {
 					fi
 				fi
 				if version_gte "20.10"; then
-					pkgs="$pkgs docker-compose-plugin docker-ce-rootless-extras$pkg_version"
-				fi
-				if version_gte "23.0"; then
-						pkgs="$pkgs docker-buildx-plugin"
+					pkgs="$pkgs docker-buildx-plugin docker-compose-plugin docker-ce-rootless-extras$pkg_version"
 				fi
 				if ! is_dry_run; then
 					set -x
@@ -701,10 +695,7 @@ do_install() {
 					fi
 				fi
 				if version_gte "20.10"; then
-					pkgs="$pkgs docker-compose-plugin docker-ce-rootless-extras$pkg_version"
-				fi
-				if version_gte "23.0"; then
-						pkgs="$pkgs docker-buildx-plugin"
+					pkgs="$pkgs docker-buildx-plugin docker-compose-plugin docker-ce-rootless-extras$pkg_version"
 				fi
 				if ! is_dry_run; then
 					set -x


thaJeztah:
CI is expected to fail currently

tianon (approved):
Ah right, that makes sense too -- it's not "required" in quite the same way on 20.10, but Docker still recommends it. :+1:

thaJeztah:
Thinking of this; I don't think I tested this yet on an rpm distro, so will need to try that as well

thaJeztah:
Was cleaning up branches, and rebased this one. Gave it a go on an RPM distro (Fedora).

First install 20.10;

```bash
docker run -it --rm quay.io/centos/centos:stream9
[root@b820b456eb96 /]# curl -fsSL get.docker.com > install.sh
[root@b820b456eb96 /]# chmod +x install.sh
[root@b820b456eb96 /]# ./install.sh --version 20.10
# Executing docker install script, commit: 6d9743e9656cc56f699a64800b098d5ea5a60020
+ sh -c 'dnf --best install -y -q dnf-plugins-core'


docker info
Client:
 Context:    default
 Debug Mode: false
 Plugins:
  app: Docker App (Docker Inc., v0.9.1-beta3)
  buildx: Docker Buildx (Docker Inc., v0.10.4-docker)
  compose: Docker Compose (Docker Inc., v2.28.1)
```

Then trying to upgrade the buildx plugin using the script from this PR;

```bash
curl -fsSL https://raw.githubusercontent.com/docker/docker-install/eb2ab892194e50aad24eaa96e9644509da0d4ded/install.sh > install.sh

Adding repo from: https://download.docker.com/linux/centos/docker-ce.repo
+ '[' stable '!=' stable ']'
+ sh -c 'dnf makecache'
CentOS Stream 9 - BaseOS                                    54 kB/s |  13 kB     00:00
CentOS Stream 9 - AppStream                                143 kB/s |  13 kB     00:00
CentOS Stream 9 - Extras packages                          146 kB/s |  14 kB     00:00
Docker CE Stable - aarch64                                  52 kB/s | 3.5 kB     00:00
Metadata cache created.
INFO: Searching repository for VERSION '20.10'
INFO: dnf list --showduplicates docker-ce | grep '20.10.*el' | tail -1 | awk '{print $2}'
+ sh -c 'dnf --best install -y -q docker-ce-20.10.24-3.el9 docker-ce-cli-20.10.24-3.el9 containerd.io docker-buildx-plugin docker-compose-plugin docker-ce-rootless-extras-20.10.24-3.el9'
Error: Transaction test error:
  file /usr/libexec/docker/cli-plugins/docker-buildx from install of docker-buildx-plugin-0:0.15.1-1.el9.aarch64 conflicts with file from package docker-ce-cli-1:20.10.24-3.el9.aarch64
```

Which.. fails; it doesn't like replacing files that were installed through existing packages.

thaJeztah:
So it looks like RPM doesn't support this scenario; from some reading up, there's 2 options to support the equivalent of `deb` doing a "replace";

- `--replacefiles` is the closest equivalent, however it's not been carried in `dnf`; https://bugzilla.redhat.com/show_bug.cgi?id=2129808
- `--force` - which is a rather big hammer

I was hoping to cleanup the script a bit with this, by combining these branches, but perhaps the better option is to hard-fail when trying to install such old versions with this script

PR #300, opened 2022-07-25
Updated the mirror repo's URL

D3-3109:
AzureChinaCloud's mirror repo is now unreachable , so i changed it to Tencent's.


D3-3109:
oops,i renamed the previous branch

PR #304, opened 2022-08-06
Changed the AzureChinaCloud's mirror repo to Tencent's

D3-3109:
AzureChinaCloud's mirror repo is now unreachable , so i changed it to Tencent's.
This new mirror repo is quite reliable.
It should be fine this time.

Signed-off-by: D3-3109 <63032814+D3-3109@users.noreply.github.com>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -66,8 +66,8 @@ case "$mirror" in
 	Aliyun)
 		DOWNLOAD_URL="https://mirrors.aliyun.com/docker-ce"
 		;;
-	AzureChinaCloud)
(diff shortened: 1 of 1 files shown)


thaJeztah (changes requested):

D3-3109:
@thaJeztah, thank you for taking the time to review my pull request and for the explanation. I understand that adding more mirrors to the list is not ideal at this time. I appreciate your suggestion of using the DOWNLOAD_URL environment variable for custom mirrors, and I will keep that in mind for future reference.

I will close this pull request as it won't be merged at this time. Thanks again for the feedback!

PR #322, opened 2022-11-07
Revert "update rootless-install to 20.10.21"

jen801:
Reverts docker/docker-install#319


thaJeztah:
Not sure why this would have to be reverted, and this is missing a DCO sign-off, so I'm closing this PR, but feel free to comment if there was a specific reason.

PR #328, opened 2022-12-24
Update rootless-install.sh

tankaras:
update rootless-install to 20.10.22


thaJeztah:
Thanks for contributing! Looks like the commit is missing a DCO sign-off.

I was working on some other changes, so I'm opening a PR to update the version.

PR #333, opened 2023-02-02
fix: add apparmor-utils to install prereqs

dvdksn:
Adds apparmor-utils as a prereq dependency

Solves issues for some Debian users on version 23.0

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -389,7 +389,7 @@ do_install() {
 	# Run setup for each distro accordingly
 	case "$lsb_dist" in
 		ubuntu|debian|raspbian)
-			pre_reqs="apt-transport-https ca-certificates curl"
+			pre_reqs="apt-transport-https ca-certificates curl apparmor-utils"
 			if ! command -v gpg > /dev/null; then
 				pre_reqs="$pre_reqs gnupg"
 			fi


vvoland (approved):

neersighted (changes requested):
I don't think this change should be accepted. `apparmor-utils` is an optional dependency we are meant to check at runtime, and this entire issue happened because containerd regressed their public API and we did not catch it. I think that a release note is likely adequate for now, as altering the semantics of the installer to work around a minor regression is likely to be painful later (and many users will not want apparmor!).

akerouanton:
`apparmor_parser` comes from package `apparmor`, not `apparmor-utils` (see [here](https://packages.debian.org/search?searchon=contents&keywords=apparmor_parser&mode=path&suite=stable&arch=any)). This package is included by default in bullseye cloud images (see [here](https://cloud.debian.org/images/cloud/bullseye/latest/debian-11-generic-amd64.json)). Thus, users have to intentionally uninstall it to face the reported bug. So I second what @neersighted said: the issue is in moby/moby or containerd, not in the install script.

thaJeztah:
I opened https://github.com/docker/docker-install/pull/335 to remove the `--no-install-recommends` flag instead.

PR #44, opened 2017-10-24
Add temporary Ubuntu Artful support

seemethere:
Adds temporary support for Ubuntu Artful while we wait to get packages
up, after we do get packages up we should delete the lines denoted `#TODO: DELETE HERE`

Signed-off-by: Eli Uriegas <eli.uriegas@docker.com>

PR #335, opened 2023-02-03
deb: remove --no-install-recommends

thaJeztah:
- supersedes, closes https://github.com/docker/docker-install/pull/333
- relates to https://github.com/docker/docker-install/pull/44
- relates to https://github.com/moby/moby/issues/44900

The installation for deb packages used a `--no-install-recommends` to skip "recommended" packages. This flag was added in ab5553e3027fdf53c5b68ef543f7437dc15c33df (https://github.com/docker/docker-install/pull/44) to work around an issue on Ubuntu 17.10 (which had aufs-tools as recommended dependency), causing the kernel to be updated.

While "recommended" packages _can_ be omitted during installation, they are a "strong" recommendation, and should really be installed by default:

> Recommends
>
> This declares a strong, but not absolute, dependency.
> The Recommends field should list packages that would be found together
> with this one in all but unusual installations.

Including the `--no-install-recommends` option could result in `apparmor_parser` binary (part of the `apparmor` package) to be missing, which would cause the daemon failing to start on systems with AppArmor support enabled:

    Error response from daemon: AppArmor enabled on system but the docker-default profile could not be loaded: running `apparmor_parser apparmor_parser --version` failed with output:
    error: exec: "apparmor_parser": executable file not found in $PATH
    Error: failed to start containers: somecontainer

This patch removes the `--no-install-recommends` so that the recommended packages are installed.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -450,7 +450,7 @@ do_install() {
 				if ! is_dry_run; then
 					set -x
 				fi
-				$sh_c "DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends $pkgs >/dev/null"
+				$sh_c "DEBIAN_FRONTEND=noninteractive apt-get install -y -qq $pkgs >/dev/null"
 				if version_gte "20.10"; then
 					# Install docker-ce-rootless-extras without "--no-install-recommends", so as to install slirp4netns when available
 					$sh_c "DEBIAN_FRONTEND=noninteractive apt-get install -y -qq docker-ce-rootless-extras${pkg_version%=} >/dev/null"


thaJeztah:
@crazy-max @vvoland PTAL

./cc @neersighted @tianon

vvoland (approved):

sam-thibault (approved):

kiorky:
As said on https://github.com/moby/moby/issues/44900#issuecomment-1415584966, i'm afraid this wont be sufficient .

kiorky:
(And moreoever, i think this may be harmful on some cases)

kiorky:
And moreover, i think this may be harmful on some cases as installing apparmor related packages during an unattended upgrade can cause unexpected behavior on systems where we are not expect it to be installed...

neersighted:
I think you misunderstand what this aims to solve -- `Recommends:` are strong dependencies that all users should install on Debian systems. They can be explicitly uninstalled by a system administrator who understands the implications of doing so -- however, that should not be **the default**.

This script is meant to be a convenience, not the correct way to install Docker CE. It should differ little from adding the repos and installing the correct packages.

There is still a regression in Moby where we cannot start with the apparmor kernel module and no userspace parser. That still should be fixed. However, this removes a misused option from a (unfortunately) common install method and aligns it back to best practices, while also mitigating the bleeding for 99% of users.

Of course refusing to start without apparmor_parser requires more than this. But this is an easy thing to do while we work on our first patch release of this branch, and should be done even if this regression did not exist.

kiorky:
I'm not sure to understand where you want to come at the end.

So i still maintain that :

- Changing an installer which will install apparmor packages where it did not in the first place can do more harm that good as it is not expected to behave this way.
- The aforementioned change will not be sufficient for a lot of scenarii where either apparmor could not be landed on the targeted system or apparmor is just not installable (docker images, not apparmor based/friendly OS which has still a generic kernel inside (EG a redhat based LXC container on a Debian based host). That what you said in a half as i think this bug that prevents the docker daemon to operate correctly should be considered critical.
- That change should have never landed on a release, that strange it wasn't  alreadycatched before.

neersighted:
`apparmor` has always been the default. It is part of the Debian and Ubuntu base systems, and when installing using **best practices** methods, `apparmor` will be installed as `Recommends:` are 'optional but default/strongly encouraged' dependencies. 

This change is merely realigning the install script to match OS conventions and best practices.

This change is not sufficient for some use cases (e.g. a CentOS userland on a Ubuntu kernel), yes. However, it does not aim, or claim to be. It solves the 99% breakage; the regression in the daemon must be addressed to fix the rare but still valid usecases like a mixed kernel/uesrland.

Finally, this was not caught as it resulted from a bump of containerd. containerd made a change in their public/stable API that they did not think was breaking, and it was poorly documented. Nobody noticed when we pulled in an updated containerd library, and QA (including my colleagues at Mirantis) did not reveal the change up to this point, as exotic workloads like mixed userland-kernel situations were not tested.

I am not sure what more you want from the contributors here. We are trying to solve this for the majority of users with a trivial fix in packaging/distribution, while we also work on solving the regression in the daemon once and for all by having our own robust checks for `apparmor_parser` instead of relying on a third-party library.

kiorky:
No, `apparmor` is not always installed as part of base. You can have only `libapparmor1` installed, specially on containers or VMs. Furthermore, installing apparmor on certain containers/vms can be tedious.

Please also reconsider the `exotic` condition as you can also add those ones which are broken but we can imagine there are a lot more:
- Using a Debian based container on a debian based host.
- Using a Debian based host where most apparmor is purged for any reason
- Using a CI on a Debian based host (or any kernel with apparmor inside) with a DIND workflow (`docker:dind` will be used in the CI pipeline)

What i am expecting is more to work on https://github.com/moby/moby/pull/44902 or something similar to let docker run as it has already done, there is no reason for the daemon not to start in this case (it can warn, but not fail).

neersighted:
These are not mutually exclusive PRs. I cannot explain that enough to you. When investigating the apparmor issue, it was noticed that this flag was (incorrectly) used in this installer, and that apparmor was not required even though it should be (unless specifically excluded e.g. by using apt pinning or `apt install docker-ce apparmor-`).

That does not mean that moby#44902 is less of a priority. It does not mean we will not merge that PR, or a similar one. However, that PR still needs to go through code review, and this PR was much easier to review and merge in the mean time.

As upstream, we can't get everything right the first time 100% of the time. We try, and we have processes in place (such as professional QA), but things can be overlooked. If you are currently broken by this, and do not want to wait for code review and a release, you can pin to the 20.10 branch, or build your own packages with that PR included.

kiorky:
Pinning to a previous release is already what we have done here to workaround, yesterday, but we prefer to have it fixed in the first place.

For both PRs, Yes, i understood they were not mutually exclusive, just wanted to emphasis on priority.

I'm not either telling you to urge and bypass any QA process at the risk to degrade things even more.
i just wanted to be sure to be understood as what i can understand: moby/moby#44900 is not priorized yet (`priority/P2` when i would have though `P0`), isn't it ?

thaJeztah:
I don't consider pinning to a version a _workaround_, but a best practice for any production scenario. Installing "latest" (including upgrades) without verifying in a staging/test environment would be dangerous to do in any scenario. There's a rather verbose disclaimer in the documentation about this as well; https://docs.docker.com/engine/install/debian/#install-using-the-convenience-script

Screenshot 2023-02-03 at 14 42 47

That said, we can still look at some of these scenarios (CentOS in a Debian environment, docker-in-docker (or LXC)), but that's out of scope for this particular repository / script.

PR #336, opened 2023-02-03
deb: simplify version-conditional installs

thaJeztah:
- follow-up to https://github.com/docker/docker-install/pull/335

Now that 0095742fb5d923f876e720731073c541f47f0f79 (https://github.com/docker/docker-install/pull/335) removed the "--no-install-recommends" option for all packages, we no longer need to special-case installation of the rootless-extras
package.

This patch moves the package together with other conditional packages for 20.10 and up. The order of these conditionals was updated to match the equivalent steps of the RPM packages.


sam-thibault (approved):

thaJeztah:
Rebased to remove the extra commit from this PR; let me get this one in 👍

PR #359, opened 2023-05-07
Assorted fixes and improvements

thaJeztah:
See individual commits for details

### fix version-compare for non-CalVer versions

The script strips leading zeros to accommodate CalVer versions, which
led to an empty when passing a SemVer(ish) version:

    VERSION=23.0.0 ./install.sh --dry-run
    ...
    ./install.sh: 118: [: Illegal number:

### fix Docker Desktop URL to prevent redirect

The URL was missing a trailing slash, which would cause a redirect:

    curl -I -s https://www.docker.com/products/docker-desktop | grep location:
    location: https://www.docker.com/products/docker-desktop/

### do not recursively chmod /etc/apt/keyrings

- relates to https://github.com/docker/docs/pull/17070

This step in the installation script was added to take non-standard umasks
into account (which is sometimes the case on some cloud providers).

The existing command was changing modes a bit too eagerly, as it recursively
set the "execute" bit, which would not only affect the /etc/apt/keyrings
directory, but also any file inside it (including files we don't own);

    mkdir -m 0700 keyrings
    touch keyrings/one.sh
    touch keyrings/docker.gpg
    
    ls -al keyrings/
    total 8
    drwx------  4 root root 128 May  7 11:43 .
    drwx------ 17 root root 544 May  7 11:43 ..
    -rw-r--r--  1 root root   0 May  7 11:43 docker.gpg
    -rw-r--r--  1 root root   0 May  7 11:43 one.sh

    chmod -R a+rx keyrings

    ls -al keyrings/
    total 0
    drwxr-xr-x  4 root root 128 May  7 11:43 .
    drwx------ 17 root root 544 May  7 11:43 ..
    -rwxr-xr-x  1 root root   0 May  7 11:43 docker.gpg
    -rwxr-xr-x  1 root root   0 May  7 11:43 one.sh

This patch changes the script to use `install`, which creates the directory
if it doesn't exist, and sets the permissions on the directory itself, without
recursing to files inside it:

    mkdir -m 0700 keyrings2
    touch keyrings2/one.sh
    touch keyrings2/docker.gpg
    
    ls -al keyrings2/
    total 8
    drwx------ 2 root root 4096 May  7 11:44 .
    drwxr-xr-x 1 root root 4096 May  7 11:44 ..
    -rw-r--r-- 1 root root    0 May  7 11:44 docker.gpg
    -rw-r--r-- 1 root root    0 May  7 11:44 one.sh

    install -m 0755 -d keyrings2/
    total 8
    drwxr-xr-x 2 root root 4096 May  7 11:44 .
    drwxr-xr-x 1 root root 4096 May  7 11:44 ..
    -rw-r--r-- 1 root root    0 May  7 11:44 docker.gpg
    -rw-r--r-- 1 root root    0 May  7 11:44 one.sh

Changing permissions of the `docker.gpg` file itself is already handled
separately through `chmod a+r /etc/apt/keyrings/docker.gpg`.

### prevent globbing when enabling/disabling rpm channels

I ran into this when I ran the script from within `/etc/yum.repos.d`:

    + sh -c 'dnf config-manager --set-disabled docker-ce-*'
    Error: No matching repo to modify: docker-ce-staging.repo.

Took me a bit to understand why the script was failing, and then realized
the shell was expanding `docker-ce-*` to filenames in my current directory;

    ls -l docker-ce-*
    -rw-r--r--. 1 root root 2027 May  6 09:32 docker-ce-staging.repo

This patch quotes the channel name to prevent globbing.

### use single-quotes for sed regex-patterns

While I haven't found cases where this caused issues, it probably doesn't
hurt to use single-quotes for these to prevent globbing.

### SLES: move opensuse_repo variable closer to where it's used

It's only used in a single location, so move it there

### remove redundant quotes when resolving versions

These were just noise :)

### add deprecation warnings for non-LTS Ubuntu versions that are EOL

Some users confuse "YY.04" versions for LTS versions, and not all users
pay close attention to versions reaching EOL. Print a warning for those
as well, because we no longer publish packages for these.

### validate CHANNEL, and mark "edge" and "nightly" deprecated

The edge channel was already deprecated; the nightly channel has not been
updated for a long time, so (for now) mark it as deprecated as well.

### validate --mirror, and sort options

Produce an error if an unknown mirror is provided. Also sort the options
in the switch.

### add --version option

Add a --version option as alternative to the $VERSION environment variable;

    ./install.sh --version 20.10
    + sh -c apt-get update -qq >/dev/null
    + sh -c DEBIAN_FRONTEND=noninteractive apt-get install -y -qq apt-transport-https ca-certificates curl gnupg >/dev/null
    debconf: delaying package configuration, since apt-utils is not installed
    + sh -c install -m 0755 -d /etc/apt/keyrings
    + sh -c curl -fsSL "https://download.docker.com/linux/ubuntu/gpg" | gpg --dearmor --yes -o /etc/apt/keyrings/docker.gpg
    + sh -c chmod a+r /etc/apt/keyrings/docker.gpg
    + sh -c echo "deb [arch=arm64 signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu jammy stable" > /etc/apt/sources.list.d/docker.list
    + sh -c apt-get update -qq >/dev/null
    INFO: Searching repository for VERSION '20.10'
    INFO: apt-cache madison docker-ce | grep '20.10' | head -1 | awk '{$1=$1};1' | cut -d' ' -f 3
    INFO: apt-cache madison docker-ce-cli | grep '20.10' | head -1 | awk '{$1=$1};1' | cut -d' ' -f 3
    + sh -c DEBIAN_FRONTEND=noninteractive apt-get install -y -qq docker-ce=5:20.10.24~3-0~ubuntu-jammy docker-ce-cli=5:20.10.24~3-0~ubuntu-jammy containerd.io docker-compose-plugin docker-ce-rootless-extras=5:20.10.24~3-0~ubuntu-jammy >/dev/null

### add --channel option

Add a --channel option as alternative to the $CHANNEL environment variable;

    ./install.sh --channel test --dry-run
    apt-get update -qq >/dev/null
    DEBIAN_FRONTEND=noninteractive apt-get install -y -qq apt-transport-https ca-certificates curl gnupg >/dev/null
    install -m 0755 -d /etc/apt/keyrings
    curl -fsSL "https://download.docker.com/linux/ubuntu/gpg" | gpg --dearmor --yes -o /etc/apt/keyrings/docker.gpg
    chmod a+r /etc/apt/keyrings/docker.gpg
    echo "deb [arch=arm64 signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu jammy test" > /etc/apt/sources.list.d/docker.list
    apt-get update -qq >/dev/null
    DEBIAN_FRONTEND=noninteractive apt-get install -y -qq docker-ce docker-ce-cli containerd.io docker-compose-plugin docker-ce-rootless-extras docker-buildx-plugin >/dev/null

### rewrite instructions

Provide more information how to use the script, its intended purpose, and
limitations of the script.


thaJeztah:
@tianon @vvoland ptal 🤗

install.sh
```diff
@@ -1,22 +1,82 @@
...
+# - The script installs dependencies and recommendations without asking for
+#   confirmation.
+# - By default, the script installs the latest stable release of Docker CLI,
+#   Docker Engine, Docker Buildx, Docker Compose, containerd, and runc. When
+#   using this script to provision a machine, this may result in unexpected
+#   major version upgrades of these packages. Always test upgrades in
+#   a test environment before deploying to your production systems.
+# - The script isn't designed to upgrade an existing Docker installation. When
+#   using the script to update an existing installation, dependencies may not be
+#   updated to the expected version, resulting in outdated versions.
```
vvoland, line 24: nit: "The script" is repeated in every bullet sentence. It could be extracted into a list header:

```suggestion
# The script:
# - Requires `root` or `sudo` privileges to run.
# - Attempts to detect your Linux distribution and version and
#   configure your package management system for you.
# - Doesn't allow you to customize most installation parameters.
# - Installs dependencies and recommendations without asking for
#   confirmation.
# - Installs the latest stable release (by default) of Docker CLI,
#   Docker Engine, Docker Buildx, Docker Compose, containerd, and runc. When
#   using this script to provision a machine, this may result in unexpected
#   major version upgrades of these packages. Always test upgrades in
#   a test environment before deploying to your production systems.
# - Isn't designed to upgrade an existing Docker installation. When
#   using the script to update an existing installation, dependencies may not be
#   updated to the expected version, resulting in outdated versions.
```

vvoland (dismissed):

thaJeztah:
@vvoland updated; PTAL 🤗

PR #383, opened 2023-11-17
Drop SLES 15-SP3 support

Prabhav-Thali:
Dropped support for SLES 15-sp3 as it is EOL. 
Updated the opensuse_repo URL as for SLES 15-SP5, container-selinux is present in the Opensuse_factory repo. And made it generic for SLES 15-sp4 and SLES 15-sp5.

Signed-off-by: Prabhav Thali [Prabhav.Thali1@ibm.com](mailto:Prabhav.Thali1@ibm.com)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -450,7 +450,7 @@ do_install() {
 			esac
 		;;
 
-		centos|rhel|sles)
+		centos|rhel)
 			if [ -z "$dist_version" ] && [ -r /etc/os-release ]; then
 				dist_version="$(. /etc/os-release && echo "$VERSION_ID")"
 			fi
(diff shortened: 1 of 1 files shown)


Prabhav-Thali:
Hi @thaJeztah,
Could you please take a look? Thanks.

thaJeztah (approved):
looks like I missed your ping 🙈 

SGTM

install.sh
```diff
@@ -450,7 +450,7 @@ do_install() {
 			esac
 		;;
 
-		centos|rhel|sles)
+		centos|rhel)
```
thaJeztah, line 453: Actually wondering now if we need this option in the switch at all 🤔 (looks like same code is also in the `*)` case 🤔

PR #386, opened 2024-01-07
Added a hint for a very usual post install step

tonivega:
I would like to add a hint for post install steps as I'm unable to remember the exact command and I'm always googling that command after using the install script.

I believe that it will be very useful for everyone to mention the exact command for giving permissions to the current user to use the docker daemon.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -301,7 +301,12 @@ echo_docker_as_nonroot() {
 	echo "         documentation for details: https://docs.docker.com/go/attack-surface/"
 	echo
 	echo "================================================================================"
-	echo
+	echo 
+	echo "Hint: You might want to grant Docker access to the current user by executing:"
+	echo 
+	echo '         sudo usermod -aG docker $USER'
+	echo 
+	echo "================================================================================"
 }
 
 # Check if this is a forked Linux distro


install.sh
```diff
@@ -301,7 +301,12 @@ echo_docker_as_nonroot() {
 	echo "         documentation for details: https://docs.docker.com/go/attack-surface/"
 	echo
 	echo "================================================================================"
-	echo
+	echo 
+	echo "Hint: You might want to grant Docker access to the current user by executing:"
+	echo 
+	echo '         sudo usermod -aG docker $USER'
+	echo 
+	echo "================================================================================"
```
AkihiroSuda, line 309: This was removed on purpose in favor of `dockerd-rootless-setuptool.sh install`:
- https://github.com/docker/docker-install/pull/215

thaJeztah: Correct; links were added for rootless, and the https://docs.docker.com/go/daemon-access/ link provides information how to run as a non-root user.

We prever using those links, as both provides relevant information on making these changes, and also outline the risks of doing so (having access to the API allows a non-root user to escalate to `root` on the host; https://docs.docker.com/go/attack-surface/).

thaJeztah:
Thanks for contributing, I'm closing this PR for reasons outlined above, but feel free to continue the conversation 👍

PR #395, opened 2024-02-10
CentOS/RHEL: Use dnf if available

### LINKED ISSUE #394: yum is no longer available by default in CentOS Stream 9
The current install script still attempts to use yum in any version of CentOS (Stream).
However yum is no longer available in recent versions (like Stream 9) and users have to install the yum package manually in advance.

kajinamit:
... instead of yum. The yum command is no longer available by default in recent versions.

Closes #394

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -559,19 +559,24 @@ do_install() {
 				echo "Packages for RHEL are currently only available for s390x."
 				exit 1
 			fi
-			if [ "$lsb_dist" = "fedora" ]; then
+
+			if command_exists dnf; then
 				pkg_manager="dnf"
 				config_manager="dnf config-manager"
 				enable_channel_flag="--set-enabled"
 				disable_channel_flag="--set-disabled"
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM - looks like a sane approach, thanks!

thaJeztah:
@neersighted PTAL


FWIW; one difference in `dnf` that  I'm aware of is that `yum install` also _updates_ packages already installed, whereas `dnf install` doesn't.

This script is not designed to do _updates_ of existing installations, but we're aware of users trying to use it for that purpose. Wondering if we should account for that 🤔

neersighted:
It looks like `--best` should trigger the 'update the packages specified in the command line' behavior that Yum had; I agree we should probably specify it.

neersighted:
I agree it's best effort; still, it's somewhat trivial to add the correct flag here. I'll amend this PR.

thaJeztah (approved):
LGTM

PR #404, opened 2024-03-20
rootless-install: Add spacing to limit merge conflicts

vvoland:
Assign `CHANNEL` variable between `STABLE_LATEST` and `TEST_LATEST` to
spread them apart so they don't appear in diff.

This allows to limit merge conflicts when there are two separate PRs
that update one of the versions.


thaJeztah (approved):
LGTM

PR #408, opened 2024-04-22
Update install.sh (add TencentCloud download url)

89ao:
add TencentCloud download url

**- What I did**
add TencentCloud download url

**- How I did it**
add TencentCloud download url(https://mirrors.cloud.tencent.com) in the install.sh

**- How to verify it**
curl -fsSL https://get.docker.com -o get-docker.sh
sudo sh get-docker.sh --mirror TencentCloud

Add the download address of TencentCloud to facilitate faster use by users within certain networks.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -139,7 +139,7 @@ case "$mirror" in
 		DOWNLOAD_URL="https://mirror.azure.cn/docker-ce"
 		;;
   	TencentCloud)
-		DOWNLOAD_URL="https://mirror.cloud.tencent.com/docker-ce"
+		DOWNLOAD_URL="https://mirrors.cloud.tencent.com/docker-ce"
 		;;
 	"")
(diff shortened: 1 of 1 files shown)


thaJeztah (changes requested):

VAllens:
> Thanks for contributing; I don't think we want to add more options to this list; the original list of mirrors were part of some partnerships (but those ended, so we may even consider removing those).
> 
> If you need to use a custom / alternative mirror, I would recommend using the env-var when executing the script.`DOWNLOAD_URL`

I was planning to add `TencentCloud` as a mirror option and pull request. 
So I found this place, and after seeing your comment, I realized that I've been using `get-docker.sh` for so long and never knew about the `DOWNLOAD_URL` environment variable.
This is great idea.

PR #413, opened 2024-05-09
rootless: update to 26.1.2

vvoland:
rootless: update to 26.1.2

PR #414, opened 2024-05-09
Makefile: Ignore SC2329

vvoland:
See https://github.com/docker/docker-install/pull/413


install.sh
```diff
@@ -179,7 +179,7 @@ version_gte() {
 	if [ -z "$VERSION" ]; then
 			return 0
 	fi
-	eval version_compare "$VERSION" "$1"
+	version_compare "$VERSION" "$1"
```
thaJeztah, line 182: Is there any change in behavior between with / without `eval` ? (LOL, I may have written this code, but I honestly don't recall why it's there 🤔 🙈)

vvoland: It seems to work fine without it:

```
bash-3.2$ export VERSION=23.0
bash-3.2$ version_gte 24.0; echo $?
1
bash-3.2$ version_gte 34.0; echo $?
1
bash-3.2$ version_gte 23.0; echo $?
0
bash-3.2$ version_gte 23.5; echo $?
1
bash-3.2$ version_gte 20.10; echo $?
0
```

vvoland: Btw, the examples in the comment seem wrong 🙈

vvoland: Looks like a mistake in d08fb9eb
```diff
-# VERSION=20.10
+# VERSION=23.0
+# version_gte 23.0  // 0 (success)
```

thaJeztah: > Looks like a mistake in [d08fb9e](https://github.com/docker/docker-install/commit/d08fb9ebee921b2a3654f719b405a5afb0522124)
> 
> ```diff
> -# VERSION=20.10
> +# VERSION=23.0
> +# version_gte 23.0  // 0 (success)
> ```

Isn't that correct though? I.e. 23.0 is "gte" (`>=`) 23.0, or didn't it work?

thaJeztah: Oh! Last line; I see it now; last line had to be updated because 21.10 would now no longer pass

thaJeztah (approved):
LGTM

PR #433, opened 2024-07-20
rootless: move versions together

thaJeztah:
- follow-up to https://github.com/docker/docker-install/pull/428

Cleaning up my local branches; I had this stashed as part of the above PR, but at the time decided to keep it for a follow-up 😂

### DIFF
--- a/rootless-install.sh
+++ b/rootless-install.sh
@@ -20,8 +20,11 @@ SCRIPT_COMMIT_SHA=UNKNOWN
 
 # This script should be run with an unprivileged user and install/setup Docker under $HOME/bin/.
 
-# The latest release is currently hard-coded.
+# latest version available in the stable channel.
 STABLE_LATEST="27.0.3"
+
+# latest version available in the test channel.
+TEST_LATEST="27.0.3"
 
 # The channel to install from:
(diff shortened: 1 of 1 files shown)


krissetto (approved):

sam-thibault (approved):

vvoland:
The separation was actually introduced on purpose https://github.com/docker/docker-install/pull/404 🙈

thaJeztah:
🤦 DOH!

thaJeztah:
Oh; but I think this still works, because we now have the comment in between them; before https://github.com/docker/docker-install/pull/404 they were next to each other, which is what the problematic part was; https://github.com/docker/docker-install/blob/c938c5c478b5d6a31537634f5ce5a8df93e8cbe2/rootless-install.sh#L31-L33

And we now have a comment in between, which probably prevents merge-conflicts because of that separation 🤞 https://github.com/docker/docker-install/blob/1ce4e39c9502b89728cdd4790a8c3895709e358d/rootless-install.sh#L23-L27

thaJeztah:
And if it doesn't work; blame me! 🫶 🫶 🫶

PR #444, opened 2024-09-18
add --no-gpg-mirror option

tiankaima:
This ensures that the GPG key is always downloaded from a trusted source.

**- What I did**

add --no-gpg-mirror option to ensure GPG key is downloaded from a trusted source

**- How I did it**

< 10 lines of code I guess...

**- How to verify it**

try `./install.sh --no-gpg-mirror --mirror Aliyun --dry-run`

should be:

```
apt-get -qq update >/dev/null
DEBIAN_FRONTEND=noninteractive apt-get -y -qq install ca-certificates curl >/dev/null
install -m 0755 -d /etc/apt/keyrings
curl -fsSL "https://download.docker.com/linux//gpg" -o /etc/apt/keyrings/docker.asc
chmod a+r /etc/apt/keyrings/docker.asc
echo "deb [arch=arm64 signed-by=/etc/apt/keyrings/docker.asc] https://mirrors.aliyun.com/docker-ce/linux/ubuntu noble stable" > /etc/apt/sources.list.d/docker.list
apt-get -qq update >/dev/null
DEBIAN_FRONTEND=noninteractive apt-get -y -qq install docker-ce docker-ce-cli containerd.io docker-compose-plugin docker-ce-rootless-extras docker-buildx-plugin >/dev/null
```

add --no-gpg-mirror option to ensure GPG key is downloaded from a trusted source


tiankaima:
This came up when we maintainers at <https://mirrors.ustc.edu.cn> discussed if we should we sync the GPG key, which we aren't as for now for security reasons.

Adding this option would give us the ability to modify our help page at <https://mirrors.ustc.edu.cn/help/docker-ce.html> for a clearer guidance.

thx.

tiankaima:
Hi @thaJeztah can I get a review or update here? seems to be buried in ur inboxes.

thaJeztah (commented):
Thanks for contributing!


So, I'm a bit on the fence; to my understanding, most users using the mirror would do so because they may not have access to the upstream (`download.docker.com`); would using the GPG key from that location work?

I wonder if instead, we should bring back the actual GPG key into this script.

We used to have the GPG key in the script itself; ISTR this was partially because the keyserver we were using was known to not be that stable, but also to spread risk; i.e. instead of having the GPG key _AND_ the packages on the same location, they would be separate. I _think_ that approach was changed in relation to (at the time) "docker enterprise", because the same script (or a variant thereof) was also used to install "docker enterprise" packages, which used a different GPG key. 

There's some discussion on this PR, which (as part of it) removed the key from the script; https://github.com/docker/docker-install/pull/39

tiankaima:
> So, I'm a bit on the fence; to my understanding, most users using the mirror would do so because they may not have access to the upstream (`download.docker.com`); would using the GPG key from that location work?

Yeah I just double checked from my laptop & VPS, `download.docker.com` is not censored (...yet), just ridiculously slow. Different ISPs might censor differently, the concern is valid, but I guess providing an alternative won't make the situation worse.
 
> I wonder if instead, we should bring back the actual GPG key into this script.

I actually recommend against changing the security model just for GFW, provided that normal user won't gain from this shift(to my understanding). For now our guide recommends `curl -fsSL https://get.docker.com -o get-docker.sh`, while [TUNA Mirror](https://mirrors.tuna.tsinghua.edu.cn/help/docker-ce/) recommend fetching from `https://raw.githubusercontent.com`, both of which are "susceptible" to censoring. This shift probably won't resolve the access issue if being the sole purpose.

tl;dr: impossible for a method to be both "secure enough" and easy to access. The current impl to mirror the whole `download.docker.com` might be considered secure enough for most, if the mirror site is trusted.

> We used to have the GPG key in the script itself; ISTR this was partially because the keyserver we were using was known to not be that stable, but also to spread risk; i.e. instead of having the GPG key _AND_ the packages on the same location, they would be separate. I _think_ that approach was changed in relation to (at the time) "docker enterprise", because the same script (or a variant thereof) was also used to install "docker enterprise" packages, which used a different GPG key.
> 
> There's some discussion on this PR, which (as part of it) removed the key from the script; #39

thx, will look into it.

I'm thinking about another solution: using `GPG_KEY_MIRROR` from env, and recommend user fetching it from another mirror site, say <https://mirrors.tuna.tsinghua.edu.cn/docker-ce/> just to spread risk.

Apologies in advance if this looks a little paranoia to you, since we are basically assuming a mirror site we host being a factor of supply chain attack.

tiankaima:
force pushed again to update the commit message

tiankaima:
Hi @thaJeztah Got a minute?

PR #448, opened 2024-09-24
Broken yum config-manager command on CentOS 7

singhsegv:
**- What I did**
`"yum config-manager" to "yum-config-manager"`

**- How I did it**

**- How to verify it**

The install script seems to be broken on CentOS 7. Installing yum-utils should enable `yum config-manager` command in later versions of the OS (haven't tested) but it doesn't do it on the CentOS 7. This PR is to point this issue out and maybe someone can test on the newers OS versions if `yum-config-manager` is good to go or not. Thanks.


thaJeztah:
Doh! Looks like I broke this in https://github.com/docker/docker-install/commit/b9824740ec430eb2ac19963cb53d54778cf542d7, which was part of;

- https://github.com/docker/docker-install/pull/441


I should warn though that we no longer support CentOS 7 (which was the last distro using yum) as it reached EOL, so no updated packages will be published, and the portions of the script that use yum may eventually be removed.

So if you're still running CentOS 7 (or 8), make sure you update to a supported distro.

thaJeztah (approved):
LGTM, thanks!

GantzGit:
Was this pushed through to the get.docker.com site? I am seeing that the site is still outputting 'yum config-manager' :

Screenshot 2024-10-08 at 3 12 51 PM

thaJeztah:
Hm... let me check; I think a deploy was triggered at the time, but wondering if something failed.

thaJeztah:
Triggered a new deploy; looks to be all good now, and at commit https://github.com/docker/docker-install/commit/6d51e2cd8c04b38e1c2237820245f4fc262aca6c

```bash
curl -fsSL get.docker.com | grep 'SCRIPT_COMMIT_SHA='
SCRIPT_COMMIT_SHA="6d51e2cd8c04b38e1c2237820245f4fc262aca6c"

curl -fsSL test.docker.com | grep 'SCRIPT_COMMIT_SHA='
SCRIPT_COMMIT_SHA="6d51e2cd8c04b38e1c2237820245f4fc262aca6c"

curl -fsSL get.docker.com/rootless | grep 'SCRIPT_COMMIT_SHA='
SCRIPT_COMMIT_SHA=6d51e2c
```

PR #458, opened 2024-12-05
rpm: always replace existing repo-configuration

thaJeztah:
The installation script's main purpose is to perform a fresh install, because it's not guaranteed to work correctly when updating an existing installation, and we print a warning if we find docker is already installed. However, it's known that the script is used to update existing installations.

The current script has some limitations for the update scenario;

- When using the script for installing from the staging domain, the existing repo configuration is not overwritten. When using `yum`, or older `dnf` versions this resulted in 2 separate repo files to be created (docker-ce.repo and docker-ce-staging.repo), which resulted in errors due to duplicate repo definitions with the same name.
- When using `dnf5`, we specified the filename to store the configuration, which resulted in an error;

        + sh -c 'dnf5 config-manager addrepo --save-filename=docker-ce.repo --from-repofile='\''https://download-stage.docker.com/linux/fedora/docker-ce-staging.repo'\'''
          File "/etc/yum.repos.d/docker-ce.repo" already exists and configures repositories with IDs "docker-ce-stable docker-ce-stable-debuginfo docker-ce-stable-source docker-ce-test docker-ce-test-debuginfo docker-ce-test-source docker-ce-nightly docker-ce-nightly-debuginfo docker-ce-nightly-source". Add "--overwrite" to overwrite.

  - On SLES, it would result in a similar error;

        + sh -c 'zypper addrepo https://download-stage.docker.com/linux/sles/docker-ce-staging.repo'
          Adding repository 'Docker CE Nightly - s390x' ......................................................................................[error]
          Repository named 'docker-ce-nightly' already exists. Please use another alias.
          Adding repository 'Docker CE Nightly - Debuginfo s390x' ............................................................................[error]
          Repository named 'docker-ce-nightly-debuginfo' already exists. Please use another alias.

This patch updates the script to overwrite / replace the existing repo config. This brings the rpm-based installations in line with the Debian approach, which unconditionally overwrites `/etc/apt/sources.list.d/docker.list`.

It's worth noting that this change will therefore overwrite any customisations made to the repo-configurations, so a warning is added for that.

**- What I did**

**- How I did it**

**- How to verify it**


install.sh
```diff
@@ -568,6 +570,7 @@ do_install() {
 					$sh_c "dnf makecache"
 				elif command_exists dnf; then
 					$sh_c "dnf -y -q --setopt=install_weak_deps=False install dnf-plugins-core"
+					$sh_c "rm -f /etc/yum.repos.d/docker-ce.repo  /etc/yum.repos.d/docker-ce-staging.repo"
```
thaJeztah, line 573: Not sure if there's  cleaner ways to do this; I couldn't find subcommands  or options in older `dnf` versions to replace existing repo-configs or to easily "reconcile" the options for each repo (without writing complex steps to generate the `baseurl` etc).

install.sh
```diff
@@ -659,18 +663,23 @@ do_install() {
...
-						WARNING!!
-						openSUSE repository (https://download.opensuse.org/repositories/security:/SELinux) will be enabled now.
-						Do you wish to continue?
-						You may press Ctrl+C now to abort this script.
-						EOF
-						( set -x; sleep 30 )
+
+				opensuse_factory_url="https://download.opensuse.org/repositories/security:/SELinux/openSUSE_Factory/"
+				if ! zypper lr -d | grep -q "${opensuse_factory_url}"; then
+					opensuse_repo="${opensuse_factory_url}security:SELinux.repo"
```
thaJeztah, line 671: For this repository, I didn't want to do a blanket "remove the file", because we have no real control over the filename (`/etc/zypp/repos.d/security_SELinux.repo`), which I _think_ is based on `security:SELinux.repo` -> `security_SELinux.repo`, but not 100% sure how much of a promise that is.

So instead of removing the file, I decided to check if a repository pointing to the given URL is already  configured, which should probably cover this situation.

vvoland (approved):

PR #466, opened 2025-01-13
Update docker support for s390x

prabhav-thali:
**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -554,6 +554,10 @@ do_install() {
 			exit 0
 			;;
 		centos|fedora|rhel)
+			if [ "$(uname -m)" = "s390x" ]; then
+				echo "Effective v27.5, please consult RHEL distro statement for s390x support."
+				exit 1
+			fi
 			repo_file_url="$DOWNLOAD_URL/linux/$lsb_dist/$REPO_FILE"
 			(
 				if ! is_dry_run; then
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -651,84 +651,6 @@ do_install() {
 			echo_docker_as_nonroot
 			exit 0
 			;;
-		sles)
-			if [ "$(uname -m)" != "s390x" ]; then
```
vvoland, line 655: Could we also print some message with an explanation?

Is it okay to drop support for installing versions that were already released for SLES?

prabhav-thali: > Is it okay to drop support for installing versions that were already released for SLES?

Yes we can drop support.

vvoland: Could we still print an error here saying that SLES packages are no longer supported?

thaJeztah: Sorry, I missed this PR!


Curious; is the plan to drop immediately, or with the next major release (docker v28)? Wondering if that would be an option, so that it can be more clearly communicated that "starting with v28, SLES on s390x is no longer supported" (or something along those lines).


> Could we still print an error here saying that SLES packages are no longer supported?

Yeah, probably could use a message added here; https://github.com/docker/docker-install/blob/24f7a3141dd6bd50ee46b019b0f4690119143627/install.sh#L469-L492

prabhav-thali: We plan to drop support immediately. When the installation is tried on SLES and RHEL (for s390x), we can add a note rather than a deprecation notice. The note will mention "Effective v27.5, please consult SLES distro statement for s390x support."

install.sh
```diff
@@ -652,82 +652,8 @@ do_install() {
...
-						pkgs="$pkgs docker-buildx-plugin"
-				fi
-				if ! is_dry_run; then
-					set -x
-				fi
-				$sh_c "zypper -q install -y $pkgs"
-			)
-			echo_docker_as_nonroot
-			exit 0
+			echo "Effective v27.5, please consult SLES distro statement for s390x support."
```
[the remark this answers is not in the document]
prabhav-thali: Yes, we don't want to allow users to install versions before v27.5

vvoland (approved):

install.sh
TAMERALKHATE3B, line 1: 👋

PR #480, opened 2025-03-31
install: Default to staging repo file for staging download URL

vvoland:
When using a staging download URL (containing "-stage" in the URL), and no explicit repo was provided automatically use the staging repo file as well.

(This only affects DNF distros)

So it's sufficient to do:

```
curl -L get.docker.com | DOWNLOAD_URL="https://download-stage.docker.com" sh -
```

instead of
```
curl -L get.docker.com | REPO_FILE="docker-ce-staging.repo" DOWNLOAD_URL="https://download-stage.docker.com" sh -
```

Signed-off-by: Paweł Gronowski <pawel.gronowski@docker.com>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -101,6 +101,11 @@ fi
 DEFAULT_REPO_FILE="docker-ce.repo"
 if [ -z "$REPO_FILE" ]; then
 	REPO_FILE="$DEFAULT_REPO_FILE"
+	# Automatically default to a staging repo fora
+	# a staging download url (download-stage.docker.com)
+	case "$DOWNLOAD_URL" in
+		*-stage*) REPO_FILE="docker-ce-staging.repo";;
+	esac
 fi
 
 mirror=''


install.sh
```diff
@@ -99,8 +99,15 @@ if [ -z "$DOWNLOAD_URL" ]; then
...
 DEFAULT_REPO_FILE="docker-ce.repo"
+DEFAULT_STAGING_REPO_FILE="docker-ce-staging.repo"
+
 if [ -z "$REPO_FILE" ]; then
 	REPO_FILE="$DEFAULT_REPO_FILE"
+    # Automatically default to a staging repo fora
+    # a staging download url (download-stage.docker.com)
+    case "$DOWNLOAD_URL" in
+        *-stage*) REPO_FILE="$DEFAULT_STAGING_REPO_FILE";;
+    esac
```
thaJeztah, line 110: Looks like the new lines accidentally used spaces for indentation (I thought we had `shfmt` running in this repo, but looks like we don't? (we should add it).

```suggestion
DEFAULT_STAGING_REPO_FILE="docker-ce-staging.repo"

if [ -z "$REPO_FILE" ]; then
	REPO_FILE="$DEFAULT_REPO_FILE"
	# Automatically default to a staging repo fora
	# a staging download url (download-stage.docker.com)
	case "$DOWNLOAD_URL" in
		*-stage*) REPO_FILE="$DEFAULT_STAGING_REPO_FILE";;
	esac
```


One thing I was looking at, but we already have some cases where we did, was to keep "internal" vars lowercase, so that it's slightly more clear which env-vars we expect users to be using for overrides. (We can't use `local` because that's not a POSIX thing).

```suggestion
default_staging_repo_file="docker-ce-staging.repo"

if [ -z "$REPO_FILE" ]; then
	REPO_FILE="$DEFAULT_REPO_FILE"
	# Automatically default to a staging repo fora
	# a staging download url (download-stage.docker.com)
	case "$DOWNLOAD_URL" in
		*-stage*) REPO_FILE="$default_staging_repo_file";;
	esac
```

Or ... we could keep it simple, because we don't use that var anywhere else;

```suggestion
if [ -z "$REPO_FILE" ]; then
	REPO_FILE="$DEFAULT_REPO_FILE"
	# Automatically default to a staging repo fora
	# a staging download url (download-stage.docker.com)
	case "$DOWNLOAD_URL" in
		*-stage*) REPO_FILE="docker-ce-staging.repo";;
	esac
```

thaJeztah (changes requested):
Left a comment (at least the tab <--> spaces should be fixed 😄 )

install.sh
thaJeztah: OH! There's a typo  as well (`fora` -> `for`)

thaJeztah (approved):
LGTM

PR #488, opened 2025-04-17
gha: Add workflow to test rootless install

vvoland:
Test the rootless install in CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,3 +23,15 @@ jobs:
       run: make shellcheck
     - name: Check distribution
       run: TEST_IMAGE=${{ matrix.os }} VERSION=${{ matrix.version }} make test
+
+  # This is a separate workflow step, because we need to check it outside of container (due to lsmod, iptables checks)
+  test-install-rootless:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -23,3 +23,12 @@ jobs:
...
     - name: Check distribution
       run: TEST_IMAGE=${{ matrix.os }} VERSION=${{ matrix.version }} make test
+
+  test-install-rootless:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+    - uses: actions/checkout@v3
+    - name: Rootless install
+      run: FORCE_ROOTLESS_INSTALL=1 ./rootless-install.sh
```
vvoland, line 33: This is a separate workflow step, because we need to check it outside of container (due to `lsmod`, `iptables` checks)

robmry (approved):

.github/workflows/ci.yml
thaJeztah: Is it worth adding a comment for that? (i.e. why we need to do separate)

vvoland: Added a comment

thaJeztah (approved):
LGTM, thanks!

Issue #509, opened 2025-07-10
On ubuntu bionic: docker-model-plugin block the install


rejeangroleau:
On ubuntu **bionic**, only since today (Thursday, July 10, 2025):
When running the usual wget of get.docker.com piped into sh,
the install script ends up failing with an error about **not being able to apt-install the docker-model-plugin package.**

Running the same apt command (mentioned in the install script error log) but **removing** the last package (docker-model-plugin) solves the issue.

But it would be nice to be able to do (as it used to work for the past years) just a simple "wget piped into sh" and succeed (**without any patching** afterwards).

It's ok if you want to just close and delete this issue,
I just wanted to mention it to someone somewhere
(since not everybobdy is using such old ubuntu versions)
and it looked like here was the best place to mention it.

I hope I did not bother anyone needlessly, in that case I am sorry about it.

neuber77:
We had this issue running GitLab runners with docker-machine and creating instances using ubuntu bionic AMI.

Docker-machine uses the "https://get.docker.com" script to install docker. With this issue each new instance that is created by docker-machine is right after deleted because of this error in docker installation:

> "error installing docker"

Running tests we found the detailed error:

> E: Unable to locate package docker-model-plugin

As a workaround (while this issue is not resolved) we fixed by setting the URL of the script to a previous version without the
"docker-model-plugin":

> --engine-install-url=https://raw.githubusercontent.com/docker/docker-install/51fb9146580c75fe07421a1451b060eea126a832/install.sh

rejeangroleau:
@vvoland Thanks for the quick reply! I will use @neuber77 solution for now (to reference a fixed previous-version commit of the script) instead of creating a PR (mostly because I am unaware if the docker-model-plugin is missing just because they are still "working on it for older distros" or if it will be permanently not-available-for-older-distros, so excluding it from the source code line that you mentioned is not a clear win for now for me).

PR #511, opened 2025-07-14
Avoid installing missing packages for version 23 (and moving them to 28.2)

rejeangroleau:
On ubuntu bionic and some other debian/versions, the docker-model-plugin package is missing. The suggestion was to just install the package named docker-buildx-plugin in the version 23 check, while moving the missing package docker-model-plugin to a new "version 28.2" check.

See: https://github.com/docker/docker-install/issues/509

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -554,7 +554,10 @@ do_install() {
 						pkgs="$pkgs docker-compose-plugin docker-ce-rootless-extras$pkg_version"
 				fi
 				if version_gte "23.0"; then
-						pkgs="$pkgs docker-buildx-plugin docker-model-plugin"
+						pkgs="$pkgs docker-buildx-plugin"
+				fi
+				if version_gte "28.2"; then
+						pkgs="$pkgs docker-model-plugin"
 				fi
 				if ! is_dry_run; then
 					set -x


install.sh
```diff
@@ -656,7 +656,20 @@ do_install() {
 					pkgs="$pkgs docker-compose-plugin docker-ce-rootless-extras$pkg_version"
 				fi
 				if version_gte "23.0"; then
-						pkgs="$pkgs docker-buildx-plugin docker-model-plugin"
+						case "$lsb_dist.$dist_version" in
```
vvoland, line 659: This seems to be inside the RPM-based block, it needs to go in:

https://github.com/docker/docker-install/blob/7040dd2bf115a359317b1de84de611aeabcb7bc2/install.sh#L549-L558

vvoland, line 659: Tbh, I think we can make our life easier and only consider `model` plugin from 28.0+ up. (buildx still 23.0)

vvoland, line 659: Also, we should avoid having to maintain this list of distros. We definitely don't want to end up having adding each new distro to the list.
We should default to installing all plugins (after a version check), and then only disable for the distros we know that are not supported.

vvoland (changes requested):
Thanks! Left some comments

install.sh
rejeangroleau: @vvoland I agree, ok, so would you like me to just replace the "new" switch case by just an additional if version_gte "28.0" that contains just the model plugin (while keeping just the buildx plugin in the "if 23" section)? I could do that. Not sure if v28 is availble for me to test right now though.

rejeangroleau: @vvoland I just move it (from line 600-something to line 557). But I can also use your "version 28" approach if you want me to.

rejeangroleau: @vvoland Ok I did just that, and I used @thaJeztah suggestion of version 27 (but if you would rather use version 28 I can do that too).

vvoland:
DCO is needed: https://github.com/docker/docker-install/pull/511/checks?check_run_id=45953829168

install.sh
vvoland: Works for me. This is a best-effort check anyway 👍🏻

thaJeztah:
Yes, changing to `version_gte 28.2` sounds good to me. I originally added it to `version_gte 23.0`  with the idea that any of those versions would be recent enough to support plugins, but did not account for some distros on those versions to not have the plugin package available. This was mostly to "opportunistically" install `docker model` because some people asked "can I have `docker model` on my machine?".

I guess it's fine to just do it by default on v28.2 (or "latest") as installing "latest version" would be the most common case for users to install using this script

thaJeztah (approved):
LGTM, thanks!

thaJeztah:
@vvoland PTAL

PR #512, opened 2025-07-18
gha: Add deploy workflow

vvoland:
Add workflow that will deploy changes to the get.docker.com and test.docker.com to replace the Jenkins pipeline

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,60 @@
+name: Deploy
+
+permissions:
+  id-token: write
+  contents: read
+
+on:
+  workflow_dispatch:
+  pull_request:
+
+jobs:
+  diff:
+    name: Diff
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Diff stable
+        run: |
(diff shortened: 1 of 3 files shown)


742764146:
这是一封自动回复邮件。已经收到您的来信，我会尽快回复。

vvoland:
Ok, fine, let's move it to internal 😅

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,60 @@
+name: Deploy
+
+permissions:
+  id-token: write
```
crazy-max, line 4: Only `deploy` job requires this permission so it should be moved under it and not global.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,60 @@
...
+  workflow_dispatch:
+  pull_request:
+
+jobs:
+  diff:
+    name: Diff
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
```
crazy-max, line 17:
```suggestion
        uses: actions/checkout@v4
```

crazy-max (approved):

Issue #124, opened 2019-06-11
Option to start daemon


vipera7:
I would like to know if it's possible to start the service at the end of the installation, because starting a service is different from a centOS or Ubuntu. What do you guys think ?

gajeshbhat:
Hey @thaJeztah . I have created a Pull request to implement this feature. https://github.com/docker/docker-install/pull/518 Can you kindly take a look? Please let me know if you have any questions. Cheers. This will help close #124  CC: @vipera7

PR #518, opened 2025-08-07
Add --start-daemon option to automatically start Docker service

gajeshbhat:
## Summary

This PR adds a new `--start-daemon` command-line option that automatically starts and enables the Docker daemon service after installation, addressing issue #124.

## Problem

It would be useful to have an option to automatically start the Docker daemon after installation, noting that service management differs between distributions like CentOS and Ubuntu. Currently, users must manually run `sudo systemctl start docker` and `sudo systemctl enable docker` after installation.

## Solution

Added a `--start-daemon` flag that:
- Detects the available service management system (systemd vs traditional)
- Uses appropriate commands for each system:
  - **systemd**: `systemctl start docker && systemctl enable docker`
  - **Traditional**: `service docker start && chkconfig docker on` (or `update-rc.d`)
- Only starts the daemon when explicitly requested (safe by default)
- Supports `--dry-run` to show what commands would be executed
- Works across all supported distributions

## Usage

```bash
# Install Docker and start the daemon
sudo sh install.sh --start-daemon

# See what would happen (dry-run)
sudo sh install.sh --start-daemon --dry-run
```
## Testing
Thoroughly tested across all supported distributions:
- Ubuntu (20.04, 22.04, 24.04)
- Debian (11, 12)
- CentOS Stream (8, 9)
- Fedora (40, 41)
- RHEL/AlmaLinux (8, 9)


gajeshbhat:
@thaJeztah Can you kindly take a look ? Do you think the feature would be useful?

install.sh
```diff
@@ -75,6 +75,16 @@ set -e
 #
 #   $ sudo sh install-docker.sh --mirror AzureChinaCloud
 #
+# --start-daemon
```
vvoland, line 78: WDYT?

```suggestion
# --autostart
```

install.sh
```diff
@@ -280,6 +294,64 @@ get_distribution() {
...
+		echo "# DRY RUN: Would start and enable Docker daemon service"
+		if has_systemd; then
+			echo "# DRY RUN: systemctl start docker"
+			echo "# DRY RUN: systemctl enable docker"
+		else
+			echo "# DRY RUN: service docker start"
+			echo "# DRY RUN: chkconfig docker on (or equivalent)"
+		fi
+		return
+	fi
```
vvoland, line 314: Instead of hardcoding the logic for try run, could we just override `sh_c` so it only prints the actual command?

thaJeztah:
Oh! This one dropped off my list; I think we actually had a conversation internally to consider making this the default so that `rpm` and `deb` packages installed using the convenience script (`get.docker.com`) would always result in a running daemon.

The current script tried to align with the general conventions / expectations of `rpm` packages (which don't start services after installation), but as the `get.docker.com` script is considered a more "opinionated" way to install the docker packages, and meant for convenience (not production scenarios), we could consider doing this by default.

I had to dig back the discussion related to it, but I think it was when we made this change in the docs, which also was changed to automatically start the daemon;

- https://github.com/docker/docs/pull/21380

We could add an option to _not_ start it if we decide to make that change (`--disable-start` or `--install-only` or something along those lines)

@vvoland WDYT?

install.sh
gajeshbhat: shorter and more readable! I will update it to use `--autostart` instead of `--start-daemon`.

gajeshbhat:
> Oh! This one dropped off my list; I think we actually had a conversation internally to consider making this the default so that `rpm` and `deb` packages installed using the convenience script (`get.docker.com`) would always result in a running daemon.
> 
> The current script tried to align with the general conventions / expectations of `rpm` packages (which don't start services after installation), but as the `get.docker.com` script is considered a more "opinionated" way to install the docker packages, and meant for convenience (not production scenarios), we could consider doing this by default.
> 
> I had to dig back the discussion related to it, but I think it was when we made this change in the docs, which also was changed to automatically start the daemon;
> 
> * [engine: minor improvements to rpm-based installation instructions docs#21380](https://github.com/docker/docs/pull/21380)
> 
> We could add an option to _not_ start it if we decide to make that change (`--disable-start` or `--install-only` or something along those lines)
> 
> @vvoland WDYT?

Thanks for adding more context. I don't have a strong opinion on whether this feature should be default/non-default. I would personally like a `--autostart` option so that the default behaviour is as currently expected, ensuring that people who don't use this option/not aware of the update are not surprised if they have an automation script(s) that pulls the latest version of this script.

vvoland:
Yeah something like `--no-autostart` would also be fine.

thaJeztah:
Heh, yeah, for context; the `get.docker.com` script has had many debates over the years; it's generally "bad practice" to just `curl | sudo sh` a script; it's risky, and even more so if the script involves trying to detect what commands to run; potentially changing system settings that were already set; it's the reason it prints a warning if it discovers that `docker` is already installed, because this may mean it's installed through other means, with different configs. So it's much better to install things "manualy" (which could mean a customised variants of the steps, but targeted at the distro you're using)

That said; it's it sure is _REALLY_ convenient to use, so the script is unlikely to go away; we consider it an opinionated way to configure things, which may involve things like this (automatically start), and with the proper warnings in place ("know what you're opting into when using the script") both in the script, the "dry run" option to check what you're about to run, and warnings in our docs, I think that's perfectly fine to do.

gajeshbhat:
> Heh, yeah, for context; the `get.docker.com` script has had many debates over the years; it's generally "bad practice" to just `curl | sudo sh` a script; it's risky, and even more so if the script involves trying to detect what commands to run; potentially changing system settings that were already set; it's the reason it prints a warning if it discovers that `docker` is already installed, because this may mean it's installed through other means, with different configs. So it's much better to install things "manualy" (which could mean a customised variants of the steps, but targeted at the distro you're using)
> 
> That said; it's it sure is _REALLY_ convenient to use, so the script is unlikely to go away; we consider it an opinionated way to configure things, which may involve things like this (automatically start), and with the proper warnings in place ("know what you're opting into when using the script") both in the script, the "dry run" option to check what you're about to run, and warnings in our docs, I think that's perfectly fine to do.

I have refactored it to make autostart the default behaviour with a `--no-autostart` opt-out flag.  CC: @vvoland

install.sh
```diff
@@ -280,6 +297,81 @@ get_distribution() {
...
+# For traditional init systems:
+#   - Uses 'service docker start' to start the service
+#   - Uses 'chkconfig docker on' (RHEL/CentOS) or 'update-rc.d docker defaults' (Debian/Ubuntu)
+#     to enable on boot
+#
+# The function respects the dry-run mode and will only print commands without
+# executing them when $sh_c is set to 'echo' (dry-run mode).
+start_docker_daemon() {
+	echo
+	echo "Starting and enabling Docker daemon service..."
```
vvoland, line 324:
```suggestion
	>&2 echo "Starting and enabling Docker daemon service..."
```

(and other echos)

install.sh
```diff
@@ -280,6 +296,76 @@ get_distribution() {
...
+		return 1
+	fi
+}
+
+start_docker_daemon() {
+	>&2 echo
+	>&2 echo "Starting and enabling Docker daemon service..."
+
+	# In container environments, systemd may be installed but not running as PID 1.
+	# We try systemctl if it exists, regardless of whether systemd is the init system.
```
vvoland: Why?

gajeshbhat: On modern RHEL/CentOS/Fedora systems (like CentOS Stream 9), the `service` command doesn't exist - only `systemctl` is available for service management. In container environments `systemctl start docker fails` (systemd not running as PID 1) and `systemctl enable docker` succeeds and creates the necessary symlinks. This ensures Docker will start automatically when the system boots normally (outside the container), while gracefully handling the container case with `|| true`. Without this approach, the script fails on CentOS Stream 9 in CI with `service: command not found` error.

vvoland: Hmm I might have misunderstood something.

Even though `systemctl enable` will work in the container environment (or any other env where systemd is not the init), it won't actually run the daemon. 

So I think there's no real point in trying to run them right?


> Without this approach, the script fails on CentOS Stream 9 in CI with `service: command not found error`.

Hmm I don't think so - we shouldn't try to call `service` command if it's not available?

https://github.com/docker/docker-install/pull/518/files#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44fR333

install.sh
```diff
@@ -280,6 +296,82 @@ get_distribution() {
...
+				$sh_c 'chkconfig docker on'
+			)
+		elif command_exists update-rc.d; then
+			# Debian/Ubuntu legacy systems
+			(
+				set -x
+				$sh_c 'update-rc.d docker defaults'
+			)
+		fi
+		if ! is_dry_run; then
```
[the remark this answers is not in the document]
thaJeztah: > Looks like Ubuntu 14.04 Trusty Tahr is in the deprecation warnings list but is not blocked from installing docker and Trusty uses Upstart with Service command. I understand changing the deprecation list is out of the scope of this PR and I can create an issue for that if you'd like. This is also an additional case that we will have to consider.

https://github.com/docker/docker-install/blob/aa7ba22a7214edb5962e7ba2cb743f0f72285b6f/install.sh#L274-L286

Yes, correct, the deprecation notice only prints the warning, and adds a 10-second delay, but doesn't _refuse_ to install. The main purpose of the script at get.docker.com is to do a fresh install of the latest version of docker on supported systems, but we're aware of some users using this script to test older (EOL) versions of docker or to run docker on EOL distros for testing. We / the script don't support those, but we chose to add the warning and delay as middle-ground to raise awareness for users that may have missed their distro version fell out of support, and for those that deliberately install on unsupported combination to inform them that it's not supported (but _may_ work).

install.sh
```diff
@@ -280,6 +296,84 @@ get_distribution() {
...
+	if [ -d /run/systemd/system ]; then
+		return 0
+	else
+		return 1
+	fi
+}
+
+start_docker_daemon() {
+	>&2 echo
+	>&2 echo "Starting and enabling Docker daemon service..."
```
thaJeztah: These should be suppressed in dry-run;

```bash
	is_dry_run || >&2 echo
	is_dry_run || >&2 echo "Starting and enabling Docker daemon service..."
```

install.sh
```diff
@@ -119,6 +131,7 @@ fi
 mirror=''
 DRY_RUN=${DRY_RUN:-}
 REPO_ONLY=${REPO_ONLY:-0}
+AUTOSTART=${AUTOSTART:-1}
```
thaJeztah: Perhaps cleaner to change this to `NO_AUTOSTART` to align with the flag, and to be more clear on what's the default.

thaJeztah:
Left some comments / thoughts; I gave it a quick go to look what it looked like with my suggestions, and opened a PR against your branch; if you merge that PR, the changes should show up in this PR, and we can review / discuss those changes;

- https://github.com/gajeshbhat/docker-install/pull/1

install.sh
```diff
@@ -280,6 +296,84 @@ get_distribution() {
...
+				>&2 echo "Using systemd to manage Docker service"
+			else
+				>&2 echo "Configuring Docker service for systemd (not running as init)"
+			fi
+		fi
+		(
+			set -x
+			# Use --now to start and enable simultaneously when systemd is running
+			if has_systemd; then
+				$sh_c 'systemctl enable --now docker'
```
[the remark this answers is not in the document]
gajeshbhat: Merged from your PR. Thanks for the suggestion!

install.sh
gajeshbhat: Merged from your PR. The messages are now properly suppressed in dry-run mode.

install.sh
```diff
@@ -280,6 +296,84 @@ get_distribution() {
...
+	if command_exists systemctl; then
+		if ! is_dry_run; then
+			if has_systemd; then
+				>&2 echo "Using systemd to manage Docker service"
+			else
+				>&2 echo "Configuring Docker service for systemd (not running as init)"
+			fi
+		fi
+		(
+			set -x
```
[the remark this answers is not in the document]
gajeshbhat: Merged from your PR. Thanks.

vvoland (approved):

gajeshbhat:
Thank you both @vvoland @thaJeztah . It was fun collaborting on this PR. My understanding of docker has definitly improved after this.Looking forward to contributing more in the future. Cheers.

PR #541, opened 2025-12-11
Fix installation on debian testing (forky)

orgads:
forky dist is not deployed, so use trixie instead.

This is similar to #327.


vvoland (approved):
Thanks!

PR #556, opened 2026-04-23
install.sh: Actually start Docker daemon via systemctl

vlivyur:
Installing on fresh Fedora 43 Server Edition using `curl -fsSL https://get.docker.com | sudo bash` results in a disabled Docker service. The installation script also outputs a meaningless list of all units during installation. It happens because it runs `sh -c systemctl enable --now docker.service` instead of `sh -c "systemctl enable --now docker.service"` - and only bare systemctl command is executed without the parameters. This commit fixes this behaviour.


vvoland (approved):
LGTM, thanks!

Issue #468, opened 2025-01-21
Please support Rocky Linux


Trikere:

BurlyLuo:
+1

AkihiroSuda:
There has been a PR to build rocky-native RPMs:
- https://github.com/docker/packaging/pull/138

Issue #373, opened 2023-07-24
Rocky Linux. get-docker.sh curl setup not yet supported.


rontown:
It appears as though the following command isn't yet supported to run on Rocky Linux. (Attempted on Rocky 9.2)

`curl -fsSL https://get.docker.com -o get-docker.sh sh get-docker.sh`

Throwing this out there to see if it will be picked up and supported added in future.
Thanks

AkihiroSuda:
Discussed here:
- https://github.com/docker/docker-ce-packaging/issues/933

AkihiroSuda:
Workaround: RPMs for CentOS Stream 9 can be downloaded from here, and they may work for RHEL/CentOS-like distros too

https://download.docker.com/linux/centos/9/x86_64/stable/Packages/

rontown:
Is this to say it will be added to the above command or a more involved workaround?

neersighted:
We do not build packages at this time. We are open to it, but we're currently working through a lot of technical debt in the Docker CE release process, so expect a timeline of months instead of days.

rontown:
Fair enough! Thanks for the update and good luck clearing the dept ;)

scaleoutsean:
@neersighted, since 1 year has passed, how does it look now - is it still several months, or several quarters?

PR #557, opened 2026-04-25
install: Use RHEL repo for Rocky Linux

### LINKED ISSUE #264: Unable to run on RockyLinux
Unable to run this script on Rocky Linux (https://rockylinux.org/) although same packages as centos, rhel or fedora can be used

neitsab:
86f94af03 added Rocky Linux as a recognized distro and thus the script sets up the repo[0]. However the repo doesn't currently contain the required packages, and thus the installation fails.

Rocky being fully compatible with RHEL, using its repo leads to a working installation. This aligns with the current advice from Rocky's own docs[1].

Fixes #264 #373 #468 #469

[0]: https://download.docker.com/linux/rocky/
[1]: https://docs.rockylinux.org/10/gemstones/containers/docker/

**- What I did**

Add a test changing the value of `$repo_file_url` when rocky is detected to use rhel repo instead.

**- How I did it**

Thinking a bit and writing as much

**- How to verify it**

By reading the code?

install: improve Rocky Linux support

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -631,7 +631,11 @@ do_install() {
 				echo "Effective v27.5, please consult RHEL distro statement for s390x support."
 				exit 1
 			fi
-			repo_file_url="$DOWNLOAD_URL/linux/$lsb_dist/$REPO_FILE"
+			if [ "$lsb_dist" = rocky ]; then
+			    repo_file_url="$DOWNLOAD_URL/linux/rhel/$REPO_FILE"
+			else
+			    repo_file_url="$DOWNLOAD_URL/linux/$lsb_dist/$REPO_FILE"
+			fi
 			(
 				if ! is_dry_run; then
 					set -x


thaJeztah:
I'm not sure we should do this; we deliberately don't handle most derivatives; `raspbian` is the only exception as we migrated that, but adding other derrivatives has proven to become a sliding slope where things "may" work, but also often just enough differences for things to subtly fail.

The plan is to provide add packages for Rocky, but not all are available yet, so wondering if we should either leave the script as-is, or adding a warning ( + delay?) to allow the user to cancel the script.

Any thoughts, @vvoland ?

neitsab:
TBF this is just a workaround so as not to leave the current situation broken; IMO the script had better state Rocky was unsupported until the repo was effectively setup, and only then 86f94af03c160ca4cbf402ca887582d843e9d5d1 would have made sense.

I drafted this PR because the [replies](https://github.com/docker/docker-install/commit/86f94af03c160ca4cbf402ca887582d843e9d5d1#commitcomment-183361763) to my comment on this commit invited to do so, but it is definitely a not-ideal workaround.

> The plan is to provide add packages for Rocky, but not all are available yet

This is indeed the root cause and the first issue I wanted to report, but I couldn't find where to do do so (my question below the aforementioned commit didn't get a reply). It would indeed make more a lot more sense to finish building up the packages if that's planned/in doing and then have 86f94af.

Leaving the script as-is results in a less-than-satisfying user experience:
- user runs the script on Rocky
- script sets up a repo but then fails downloading the packages from it
- user thinks official Docker convenience script is broken
- user opens an issue.

That's basically me yesterday, and might generate a bit of useless fuss.

vvoland (changes requested):
This will break `docker sbx` installation which supports Rocky 8: https://github.com/docker/sbx-releases

Official Rocky packages for Docker Engine/CLI are in the works and should be available soon, so I think it's just not worth the time to bother with such workarounds.

neitsab:
> Official Rocky packages for Docker Engine/CLI are in the works and should be available soon

That's great to here, but until then I still think https://github.com/docker/docker-install/commit/86f94af03c160ca4cbf402ca887582d843e9d5d1 should be reverted to avoid the making the install script fail due to the absence of packages in the added repo.

Feel free to close this PR then :-)

PR #558, opened 2026-05-01
rootless-install: use docker-29.4.2-2.tgz for version 29.4.2

vvoland:
The static binary tarball for Docker 29.4.2 was re-released with a "-2" suffix. Add a static_release_version() helper that maps version numbers to their actual tarball name, so that 29.4.2 resolves to docker-29.4.2-2.tgz instead of docker-29.4.2.tgz.

**- What I did**

**- How I did it**

**- How to verify it**


thaJeztah (approved):
LGTM

PR #560, opened 2026-05-19
Use deb822-style APT repository configuration

### LINKED ISSUE #559: Debian/Ubuntu installation should use deb822 .sources format
The apt package manager is moving away from the legacy `.list` format towards the new `.sources` format for repository configuration. The install script currently configures repositories using the legacy format. It should be updated to use the newer format, at least on systems that support it.

> APT is moving to a different format for configuring where it downloads packages from. The files /etc/apt/sources.list and *.list files in /etc/apt/sources.list.d/ are replaced by files still in that directory but with names ending in .sources, using the new, more readable (deb822 style) format. For details see [sources.list(5)](https://manpages.debian.org/trixie/sources.list(5)). Examples of APT configurations in these notes will be given in the new deb822 format.

Example .sources file for Docker repo:
```
Types: deb
URIs: https://download.docker.com/linux/debian
Suites: trixie
Components: stable
Signed-By: /etc/apt/keyrings/docker.asc
```

merlinz01:
Fixes #559 

**- What I did**

Refactor Docker repository setup to support new deb822 format and provide warnings for existing configurations.

**- How I did it**

Typing on a keyboard, I guess.

**- How to verify it**

Install on recent Debian or Ubuntu. `/etc/apt/sources.list.d/debian.sources` should contain something like this:

```
Types: deb
URIs: https://download.docker.com/linux/debian
Suites: trixie
Components: stable
Architectures: amd64
Signed-By: /etc/apt/keyrings/docker.asc
```

Use deb822-style APT repository configuration on Debian/Ubuntu versions that support it.

chipmunk

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -556,9 +556,27 @@ do_install() {
 	# Run setup for each distro accordingly
 	case "$lsb_dist" in
 		ubuntu|debian|raspbian)
+			use_deb822=true
+			case "$lsb_dist.$dist_version" in
+				debian.jessie|ubuntu.trusty)
+					use_deb822=false
+					;;
+			esac
 			pre_reqs="ca-certificates curl"
-			apt_repo="deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] $DOWNLOAD_URL/linux/$lsb_dist $dist_version $CHANNEL"
 			(
+				if [ "$use_deb822" = true ]; then
+					if [ -f /etc/apt/sources.list.d/docker.list ]; then
+						echo
+						echo "# WARNING: An existing Docker APT repository file using the sources.list format was found at /etc/apt/sources.list.d/docker.list."
+						echo "# Please remove this file to avoid conflicting repository settings."
+						echo
+					fi
+					apt_repo="Types: deb\nURIs: $DOWNLOAD_URL/linux/$lsb_dist\nSuites: $dist_version\nComponents: $CHANNEL\nArchitectures: $(dpkg --print-architecture)\nSigned-By: /etc/apt/keyrings/docker.asc"
+					apt_repo_file="/etc/apt/sources.list.d/docker.sources"
+				else
+					apt_repo="deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] $DOWNLOAD_URL/linux/$lsb_dist $dist_version $CHANNEL"
+					apt_repo_file="/etc/apt/sources.list.d/docker.list"
+				fi
 				if ! is_dry_run; then
 					set -x
 				fi
@@ -567,7 +585,7 @@ do_install() {
 				$sh_c 'install -m 0755 -d /etc/apt/keyrings'
 				$sh_c "curl -fsSL \"$DOWNLOAD_URL/linux/$lsb_dist/gpg\" -o /etc/apt/keyrings/docker.asc"
 				$sh_c "chmod a+r /etc/apt/keyrings/docker.asc"
-				$sh_c "echo \"$apt_repo\" > /etc/apt/sources.list.d/docker.list"
+				$sh_c "echo \"$apt_repo\" > $apt_repo_file"
 				$sh_c 'apt-get -qq update >/dev/null'
 			)
 


install.sh
```diff
@@ -556,9 +556,27 @@ do_install() {
 	# Run setup for each distro accordingly
 	case "$lsb_dist" in
 		ubuntu|debian|raspbian)
+			use_deb822=true
+			case "$lsb_dist.$dist_version" in
+				debian.jessie|ubuntu.trusty)
+					use_deb822=false
+					;;
+			esac
```
thaJeztah, line 564: I think we already make these old distros a hard failure, so we can probably skip this check; older distros are no longer supported by the script, so it's fine if things would fail.

install.sh
```diff
@@ -556,9 +556,27 @@ do_install() {
...
 			pre_reqs="ca-certificates curl"
-			apt_repo="deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] $DOWNLOAD_URL/linux/$lsb_dist $dist_version $CHANNEL"
 			(
+				if [ "$use_deb822" = true ]; then
+					if [ -f /etc/apt/sources.list.d/docker.list ]; then
+						echo
+						echo "# WARNING: An existing Docker APT repository file using the sources.list format was found at /etc/apt/sources.list.d/docker.list."
+						echo "# Please remove this file to avoid conflicting repository settings."
+						echo
+					fi
```
thaJeztah, line 573: We already have a warning if the script detects that docker is installed already (as its main purpose is for fresh installs, not upgrades); honestly, I think we can keep it simple; assume the `docker.list` is created either by this script, or manually, but for the docker installation, so just remove it;

```bash
if [ -f /etc/apt/sources.list.d/docker.list ]; then
	rm -f /etc/apt/sources.list.d/docker.list
fi
```

thaJeztah (commented):
thanks! left some quick comments after glancing over.

install.sh
merlinz01: I don't see any hard failure here. Is that in some other component?

```sh
        # Print deprecation warnings for distro versions that recently reached EOL,
        # but may still be commonly used (especially LTS versions).
        case "$lsb_dist.$dist_version" in
                centos.8|centos.7|rhel.7)
                        deprecation_notice "$lsb_dist" "$dist_version"
                        ;;
                debian.buster|debian.stretch|debian.jessie)
                        deprecation_notice "$lsb_dist" "$dist_version"
                        ;;
                raspbian.buster|raspbian.stretch|raspbian.jessie)
                        deprecation_notice "$lsb_dist" "$dist_version"
                        ;;
                ubuntu.focal|ubuntu.bionic|ubuntu.xenial|ubuntu.trusty)
                        deprecation_notice "$lsb_dist" "$dist_version"
                        ;;
                ubuntu.oracular|ubuntu.mantic|ubuntu.lunar|ubuntu.kinetic|ubuntu.impish|ubuntu.hirsute|ubuntu.groovy|ubuntu.eoan|ubuntu.disco|ubuntu.cosmic)
                        deprecation_notice "$lsb_dist" "$dist_version"
                        ;;
                fedora.*)
                        if [ "$dist_version" -lt 41 ]; then
                                deprecation_notice "$lsb_dist" "$dist_version"
                        fi
                        ;;
        esac
```

thaJeztah: Oh, you're right; it's not a hard failure (we may have in the past), but it sleeps for 10 seconds and a warning; https://github.com/docker/docker-install/blob/2687d91ddeb3bd6aeae37a90947761efdee87030/install.sh#L273

Regardless, we should prefer keeping the script focused on what's still supported, and try to avoid complexity for things no longer supported; users can still download older versions of the script, or (better) create their own implementation targeted at their situation.

merlinz01: So should I change it to a hard fail for those distro versions?

install.sh
merlinz01: Done.

install.sh
thaJeztah: It's probably fine to keep it a warning + wait for now; it's effectively saying "this may work, but we don't test or support"; for some distros that means they were only recently deprecated, so having a hard failure may be too aggressive.

PR #564, opened 2026-06-12
scripts/get-version: Select stable release for test when newer

vvoland:
The release list is already ordered newest first, so the test channel only needs to consider releases that are either prereleases or the latest stable release.
This keeps test from resolving to an older prerelease after a newer stable release is published.


thaJeztah (approved):
Thanks! I'm horrible at jq, so I trust you on this one 😅 

"LGTM"

vvoland:
It works for the current versions at least 😅

